PR #1, opened 2023-07-25
Multiple functions support 

ryan-berger:


oflatt:
Love this PR, basically what I was going to do next
Does every function guaranteed end with one return? There's no control flow right?
If so, we don't need a return effect, we just need one expression at the end of a function to return

ryan-berger:
@oflatt My idea is that Ret should always be followed End because every function should implicitly return Void, and it is easier to add an End and fix up the list than it is to figure control flow out and go searching for Return. It also makes it so that you don't have something like: `(Ret FunctionBody ReturnValue)`. Trying to keep End as cons essentially.

PR #2, opened 2023-07-29
Fix clippy lint

ezrosent:
Wanted to start experimenting with some things, noticed that lib.rs wasn't building. I think this is what you meant?


oflatt:
It's not building? The tests pass in CI

ezrosent:
Ahhh. It does build! I just had clippy lints called out as failures in my setup. This fixes that lint.

PR #4, opened 2023-08-02
Structured cfg translation

oflatt:
This PR adds structured cfg translation with a few small tests.
It doesn't yet connect it to the optimizer by translating into an egglog encoding.


src/cfg/to_structured.rs
```diff
@@ -0,0 +1,239 @@
...
+        StructuredCfgBuilder {
+            context: vec![],
+            postorder,
+            dominators,
+            cfg,
+        }
+    }
+
+    fn convert_structured(&mut self) -> Result<StructuredFunction, EggCCError> {
+        self.check_reducible()?;
```
ryan-berger, line 44: Not that I care too much, but is there a better way of doing this where as you build your tree you find an irreducible block and return an Err? Just thinking in the "parse don't validate" vain of things.

oflatt: I thought about this and I wasn't sure- however the check is pretty simple luckily, so I implemented it.

ajpal (approved):

src/cfg/to_structured.rs
```diff
@@ -0,0 +1,239 @@
...
+            .graph
+            .edges_directed(node, petgraph::Direction::Outgoing)
+            .collect::<Vec<_>>();
+        match merge_nodes.as_slice() {
+            [] => {
+                StructuredBlock::Sequence(vec![
+                    StructuredBlock::Basic(Box::new(self.cfg.graph[node].clone())),
+                    match edges.as_slice() {
+                        [] => {
+                            panic!("handled above");
```
ajpal, line 95: Handled where? What does this mean?

PR #6, opened 2023-08-06
add in hand optimized bril, add support for fn args

egg-fnatic:
I've added support for functions with arguments, although I'm not entirely certain it is sound (looking at you ReturnVal).

So far the tests are just making sure that everything passes through okay, but I've left comments on the top of each test to show the optimized version of each for when we actually attempt to optimize otherwise we will constantly fail tests

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,7 +81,7 @@ impl Optimizer {
         let ssa_prog: AbstractProgram = serde_json::from_str(&ssa_output).unwrap();
 
         let program = Program::try_from(ssa_prog)?;
-        return Ok(program)
+        Ok(program)
     }
 
     pub fn fresh(&mut self) -> String {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,10 +60,7 @@ impl Default for Optimizer {
 }
 
 impl Optimizer {
-    pub fn parse_and_optimize(
-        &mut self,
-        program: &str,
-    ) -> Result<Program, EggCCError> {
+    pub fn parse_and_optimize(&mut self, program: &str) -> Result<Program, EggCCError> {
         let parsed = Self::parse_bril(program)?;
         eprintln!("Parsed program: {}", parsed);
         let res = self.optimize(&parsed)?;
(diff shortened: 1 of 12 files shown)


oflatt:
Great work-
Right now optimized version is in a comment
Could you add the optimized versions as bril code in another file? You could just add like a `.optimized.bril` extension.
For now they won't pass (the optimized version won't match the target exactly) so have a list of tests that shouldn't pass or something like that.

tests/reassociate_mul.bril
```diff
@@ -0,0 +1,15 @@
+# (x * (5 * (y * (6 * z)))) -> (x * (y * (z * 30)))
+@mul_reassoc (x: int, y: int, z: int): int {
```
oflatt, line 2: might be a problem if main doesn't use this function- might get dead code eliminated in the future

src/lib.rs
```diff
@@ -61,14 +60,17 @@ impl Default for Optimizer {
 }
 
 impl Optimizer {
-    pub fn parse_and_optimize(&mut self, program: &str) -> Result<Program, EggCCError> {
+    pub fn parse_and_optimize(
+        &mut self,
+        program: &str,
+    ) -> Result<Program, Box<dyn std::error::Error>> {
```
oflatt, line 66: I don't think this should be a `std::error::Error`. The benefit of the `EggCCError` is we control the output- we want to output good error messages that have meaning to the user of eggcc, not random rust errors

tests/reassociate_mul.bril
egg-fnatic: We might similarly have an issue if we do call the function and an inliner determines it to be inlineable :)

I lean towards not doing deadcode on functions for now, unless we're pushing for codegen size as a metric of success which I don't think we should/are.

oflatt: Good point, lets just deal with this if it becomes a problem

egg-fnatic:
@oflatt We should probably just add in the hand-optimized bril and kill the rest. I'll see if I can get this done soonish as it isn't a large priority.

The test cases that I added for hand-optimized bril will help us from getting any sort of blowup in associativity/commutativity which will be nice to have

oflatt:
sounds good to me

oflatt:
@ryan-berger I think these benchmarks are still valuable, esp. because I implemented reassociation for addition
Would you open a PR for these benchmarks? You can even take a crack at optimizing them if you want

oflatt:
closing since this is quite out of date

PR #8, opened 2023-08-09
Example for summing the first n positive numbers

wilcoxjay:
This PR adds bril code for summing the first n positive numbers in a few
different ways: with a straightforward loop, and with two different closed
forms.

One day hopefully we can optimize one into the other!


oflatt:
We can merge this if you just put it in a folder called `failing`

PR #9, opened 2023-08-14
Round trip optimization of structured Bril

oflatt:


src/cfg/structured.rs
```diff
@@ -1,17 +1,17 @@
...
+#[derive(Debug, PartialEq, Clone)]
 pub enum StructuredBlock {
     // Variable, then and else blocks
     Ite(String, Box<StructuredBlock>, Box<StructuredBlock>),
     Loop(Box<StructuredBlock>),
     Block(Box<StructuredBlock>),
     Sequence(Vec<StructuredBlock>),
     // how many layers of blocks / loops to break out of
-    Break(usize),
+    Break(i64),
```
ajpal, line 14: why did this need to change?

src/cfg/structured.rs
```diff
@@ -33,7 +33,21 @@ impl Display for StructuredProgram {
     }
 }
 
-#[derive(Debug)]
+impl StructuredProgram {
+    pub fn to_program(&self) -> Program {
```
ajpal, line 37: Can't comment on program.rs because it's not in the diff, but should `Program` be renamed to `UnstructuredProgram` for clarity?

src/cfg/structured.rs
```diff
@@ -91,3 +105,131 @@ impl StructuredBlock {
         }
     }
 }
+
+impl StructuredFunction {
+    pub fn to_function(&self) -> Function {
```
ajpal, line 110: same here- `UnstructuredFunction`?
Maybe a tradeoff with verbosity, but to me it feels confusing to specify structured but not unstructured (esp because the `Vec<StructuredFunctions>` field in `StructuredProgram` is named `functions`, not `structured_functions` or something)

src/cfg/structured.rs
```diff
@@ -91,3 +105,131 @@ impl StructuredBlock {
...
+            pos: None,
+            return_type: None,
+        }
+    }
+}
+
+pub struct StructuredCfgBuilder {
+    code: Vec<Code>,
+    scopes: Vec<String>,
+    name_counter: usize,
```
ajpal, line 131: `block_counter`?

src/cfg/structured.rs
```diff
@@ -91,3 +105,131 @@ impl StructuredBlock {
...
+}
+
+impl StructuredCfgBuilder {
+    fn fresh(&mut self) -> String {
+        self.name_counter += 1;
+        format!("sblock___{}", self.name_counter - 1)
+    }
+
+    fn scope_break_to(&self, num: i64) -> String {
```
ajpal, line 140: what is `num` an why must it be greater than 0? Can you add doc comments?

src/cfg/structured.rs
```diff
@@ -91,3 +105,131 @@ impl StructuredBlock {
...
+            args: self.args.clone(),
+            instrs: builder.code,
+            pos: None,
+            return_type: None,
+        }
+    }
+}
+
+pub struct StructuredCfgBuilder {
+    code: Vec<Code>,
```
ajpal, line 129: nit: can the name of this be plural to indicate that it's a vec? I found it confusing later when you do `builder.code.extend...` because the name `code` seems singular

src/cfg/structured.rs
```diff
@@ -91,3 +105,131 @@ impl StructuredBlock {
...
+    scopes: Vec<String>,
+    name_counter: usize,
+}
+
+impl StructuredCfgBuilder {
+    fn fresh(&mut self) -> String {
```
ajpal, line 135: nit: `fresh_block`? or `fresh_block_name`?

src/cfg/structured.rs
```diff
@@ -91,3 +105,131 @@ impl StructuredBlock {
...
+        self.scopes[self.scopes.len() - num as usize].clone()
+    }
+}
+
+impl StructuredBlock {
+    pub(crate) fn to_code(&self, builder: &mut StructuredCfgBuilder) {
+        match self {
+            StructuredBlock::Basic(block) => builder.code.extend(block.to_code()),
+            StructuredBlock::Block(block) => {
+                let block_end_name = builder.fresh();
```
ajpal, line 151: what does `end` mean here?

src/cfg/structured.rs
```diff
@@ -91,3 +105,131 @@ impl StructuredBlock {
...
+                builder.scopes.push(block_end_name.clone());
+                block.to_code(builder);
+                builder.code.push(Code::Label {
+                    label: block_end_name.clone(),
+                    pos: None,
+                });
+                assert!(builder.scopes.pop().unwrap() == block_end_name);
```
ajpal, line 158: intentional to remove the scope you just added? why?

src/cfg/structured.rs
```diff
@@ -91,3 +105,131 @@ impl StructuredBlock {
...
+                    builder.code.push(Code::Instruction(Instruction::Effect {
+                        op: EffectOps::Jump,
+                        args: vec![],
+                        funcs: vec![],
+                        labels: vec![builder.scope_break_to(*num)],
+                        pos: None,
+                    }));
+                }
+            }
+            StructuredBlock::Ite(cond, block1, block2) => {
```
ajpal, line 171: nit: `then_block` and `else_block` instead of `1` and `2`

src/conversions.rs
```diff
@@ -1,23 +1,27 @@
...
+    },
+    Optimizer,
+};
+use bril_rs::{EffectOps, Instruction, Literal, Type, ValueOps};
 use egglog::ast::{Expr, Symbol};
 use ordered_float::OrderedFloat;
 
 impl Optimizer {
-    pub(crate) fn expr_to_func(&mut self, expr: Expr) -> Function {
+    pub(crate) fn expr_to_func(&mut self, expr: Expr) -> StructuredFunction {
```
ajpal, line 15: `expr_to_structured_func`?

src/conversions.rs
```diff
@@ -30,77 +34,190 @@ impl Optimizer {
...
+                }
+                ("Sequence", [block, rest]) => StructuredBlock::Sequence(vec![
+                    self.expr_to_structured_block(block),
+                    self.expr_to_structured_block(rest),
+                ]),
+                ("Break", [n]) => {
+                    if let Expr::Lit(egglog::ast::Literal::Int(n)) = n {
+                        StructuredBlock::Break(*n)
+                    } else {
+                        panic!("expected int literal for break");
```
ajpal, line 62: here, and elsewhere: might be helpful to add more info for debugging, like "expected __, got __"

src/conversions.rs
```diff
@@ -30,77 +34,190 @@ impl Optimizer {
...
+                    }
+                }
+                _ => panic!("expected 2 args in expr_to_basic_block"),
+            }
+        } else {
+            panic!("expected call in expr_to_basic_block");
+        }
+    }
+
+    fn codelist_to_vec_helper(expr: &Expr, res: &mut Vec<Expr>) {
```
ajpal, line 117: is there a reason this is a void function with a mutable reference instead of returning the vec?

src/conversions.rs
```diff
@@ -111,86 +228,195 @@ impl Optimizer {
...
+        Expr::Call(
+            "Func".into(),
+            vec![
+                Expr::Lit(egglog::ast::Literal::String(func.name.clone().into())),
+                self.convert_structured_block(&func.block),
+            ],
+        )
+    }
+
+    pub(crate) fn convert_structured_block(&mut self, structured_block: &StructuredBlock) -> Expr {
```
ajpal, line 273: `structured_block_to_expr` (inverse of `expr_to_structured_block` above)

src/conversions.rs
```diff
@@ -111,86 +228,195 @@ impl Optimizer {
...
 
+    pub(crate) fn string_to_expr(&self, string: String) -> Expr {
+        Expr::Lit(egglog::ast::Literal::String(string.into()))
+    }
+
+    pub(crate) fn string_to_var_encoding(&self, string: String) -> Expr {
+        Expr::Call("Var".into(), vec![self.string_to_expr(string)])
+    }
+
+    /*pub(crate) fn string_to_var(&self, string: String) -> Expr {
```
ajpal, line 329: nit: remove if unneeded

src/conversions.rs
```diff
@@ -111,86 +228,195 @@ impl Optimizer {
...
+            Instruction::Constant {
+                dest,
+                op: _op,
+                value,
+                pos: _pos,
+                const_type,
+            } => (
+                dest.clone(),
+                match value {
+                    Literal::Int(int) => Expr::Call(
```
ajpal, line 370: Why are these `Call` not `Lit`?

src/conversions.rs
```diff
@@ -111,86 +228,195 @@ impl Optimizer {
...
+                        "Int".into(),
+                        vec![
+                            self.type_to_expr(const_type),
+                            Expr::Lit(egglog::ast::Literal::Int(*int)),
+                        ],
+                    ),
+                    Literal::Bool(bool) => {
+                        if *bool {
+                            Expr::Call("True".into(), vec![self.type_to_expr(const_type)])
```
ajpal, line 379: Why does true have an argument and false doesn't? a comment would be helpful here

src/conversions.rs
```diff
@@ -111,86 +228,195 @@ impl Optimizer {
...
+            Instruction::Value {
+                dest,
+                args,
+                funcs,
+                op,
+                labels: _labels,
+                pos: _pos,
+                op_type,
+            } => {
+                assert!(funcs.is_empty());
```
ajpal, line 409: why is funcs guaranteed empty here? can you add a comment?

src/conversions.rs
```diff
@@ -370,4 +515,48 @@ impl Optimizer {
             _ => panic!("expected type literal. got: {}", expr),
         }
     }
+
+    pub(crate) fn pretty_print_expr(expr: &Expr) -> String {
+        Self::pretty_print_expr_helper(expr, 0)
+    }
+    pub(crate) fn pretty_print_expr_helper(expr: &Expr, indent: usize) -> String {
```
ajpal, line 522: nit: more descriptive name than `helper`. `with_acc` or something?

src/cfg/structured.rs
oflatt: Program is a bril type, I could make an alias but that would be confusing. Otherwise this would be a good idea

src/cfg/structured.rs
oflatt: renamed to fresh_block_name_count

src/conversions.rs
oflatt: We have to wrap them up so the egglog typechecker is happy

src/conversions.rs
oflatt: Great catch, that's a bug. We pass the type to egglog in the encoding.

src/conversions.rs
oflatt: I think valid bril code should have funcs be empty for values, added a comment

src/cfg/structured.rs
oflatt: Added a comment (:

src/cfg/structured.rs
oflatt: Good catch, dumb reason I changed it back

src/cfg/structured.rs
oflatt: added comment- we pop the scope after recurring and thus finishing the block

src/conversions.rs
oflatt: I usually add more info as I go (hopefully we never see this)

src/conversions.rs
oflatt: the recursion would make this expensive- we could convert to linked list but it is slightly less performant

PR #11, opened 2023-08-18
port bril tests to pure Rust

egg-fnatic:


tests/files.rs
```diff
@@ -89,8 +90,61 @@ fn generate_tests(glob: &str) -> Vec<Trial> {
...
+
+fn generate_interp_tests(glob: &str) -> Vec<Trial> {
+    glob::glob(glob).unwrap().map(make_interp_test).collect()
+}
+
 fn main() {
     let args = libtest_mimic::Arguments::from_args();
-    let tests = generate_tests("tests/**/*.bril");
+    let mut tests = generate_tests("tests/{small,snapshots}/*.bril");
+    tests.append(&mut generate_interp_tests("tests/brils/**/*.bril"));
```
oflatt, line 148: Instead of making a new loop that generates tests, I would prefer that you use the existing `Run` struct to generate your tests

egg-fnatic: Sure. Should have that soon

oflatt (approved):
lgtm, next PR we are going to actually run the interpreter on the optimized code

PR #13, opened 2023-08-21
Omit unnecessary breaks

ajpal:
See 7.1 in the Relooper paper


src/cfg/to_structured.rs
```diff
@@ -135,46 +156,69 @@ impl<'a> StructuredCfgBuilder<'a> {
...
                 }
             }
         } else if self.is_backward_edge(source, target) || self.is_merge_node(target) {
             let index = self.context_index(self.cfg.graph[target].name.clone());
-            StructuredBlock::Break(index)
+            // Optimization: If target label immediately follows the hole in context, omit the Br
+            assert!(!self.context.is_empty(), "context should not be empty");
+            let top_context = self.context.last().unwrap();
+
+            if let Some(fallthrough_label) = &top_context.fallthrough {
```
oflatt, line 204: use `if let Some(true) = fallthrough.map(|fallthrough_label| fallthrough_label == target_label` or something similar

src/cfg/to_structured.rs
```diff
@@ -76,20 +84,30 @@ impl<'a> StructuredCfgBuilder<'a> {
         self.node_within(node, merge_nodes)
     }
 
-    fn node_within(&mut self, node: NodeIndex, merge_nodes: Vec<NodeIndex>) -> StructuredBlock {
+    fn node_within(
+        &mut self,
+        node: NodeIndex,
+        merge_nodes: Vec<NodeIndex>,
+    ) -> Option<StructuredBlock> {
```
oflatt, line 91: You don't need an option here- see next comment

src/cfg/to_structured.rs
```diff
@@ -76,20 +84,30 @@ impl<'a> StructuredCfgBuilder<'a> {
...
             .edges_directed(node, petgraph::Direction::Outgoing)
             .collect::<Vec<_>>();
+        assert!(!edges.is_empty(), "only exit nodes should have empty edges");
+
         match merge_nodes.as_slice() {
             [] => {
-                StructuredBlock::Sequence(vec![
-                    StructuredBlock::Basic(Box::new(self.cfg.graph[node].clone())),
+                let v = vec![
+                    Some(StructuredBlock::Basic(Box::new(
```
oflatt, line 108: When do_branch returns None, on line 116 you can return this 
that way this function doesn't return an option

oflatt (changes requested):

oflatt:
I think you need to run `cargo insta review`

oflatt (approved):
Great work!

PR #16, opened 2023-08-22
Nightly script updates

ajpal:
Pretty bare-bones atm, but we should merge so that we can keep adding features
https://nightly.cs.washington.edu/reports/eggcc/1692739633%3Anightly%3Anightly%3A3e58a5ccad/


oflatt (approved):
great!

PR #17, opened 2023-08-23
Fix several bugs in the dumb encoding

oflatt:
This PR makes the dumb encoding run on more tests in our test suite with the interpreter. There were lots of minor bugs. Also, it turns off SSA conversion for now, since we don't handle it well yet (what does it mean in the context of a structured program)?

Below are some comments about the changes


src/cfg/mod.rs
```diff
@@ -255,8 +255,26 @@ pub(crate) fn to_cfg(func: &Function) -> Cfg {
             Code::Instruction(i) => block.push(i.clone()),
         }
     }
+    // last block can implicity return, add an edge for that case
```
oflatt, line 258: fixed a bug here in Eli's cfg converter

src/conversions.rs
```diff
@@ -186,6 +214,21 @@ impl Optimizer {
                         pos: None,
                     });
                 }
+                "alloc" => {
```
oflatt, line 217: to deal with memory I added alloc here

src/lib.rs
```diff
@@ -106,16 +110,34 @@ impl Optimizer {
...
+        );
+
+        let res: Program = serde_json::from_str(&dead_code_optimized)
+            .map_err(|err| EggCCError::ConversionError(err.to_string()))?;
+        */
+
+        let prog = Program::try_from(abstract_prog)
+            .map_err(|err| EggCCError::ConversionError(err.to_string()))?;
+
+        Optimizer::check_for_uninitialized_vars(&prog)?;
```
oflatt, line 138: Throw an error for uninitialized variables.
Really just checks for  `__undefined` in the ssa converted program. For now, does nothing.

src/lib.rs
```diff
@@ -248,9 +271,17 @@ impl Optimizer {
...
+
+        (datatype ArgList
+            (ArgCons Argument ArgList)
+            (ArgNil))
+
         (datatype Function
-          ;; name and body
-          (Func String StructuredBlock))
+          ;; name, arguments, and body
+          (Func String ArgList StructuredBlock))
```
oflatt, line 284: support function arguments and passing them on the command line

src/main.rs
```diff
@@ -42,8 +45,14 @@ fn main() {
         input = std::fs::read_to_string(args.file).unwrap();
     }
 
+    let bril_args = if args.bril_args.is_empty() {
```
oflatt, line 48: Here we either get bril arguments from the command line, and if not we default to the arguments in the bril comment

tests/files.rs
```diff
@@ -84,13 +85,7 @@ fn generate_tests(glob: &str) -> Vec<Trial> {
...
 
         // TODO optimizer doesn't support these yet
-        let banned = [
-            "diamond",
-            "fib",
-            "queens_func",
-            "unstructured",
-            "implicit_return",
-        ];
+        let banned = ["queens_func", "unstructured", "implicit_return"];
```
oflatt, line 88: still need to support calling functions

oflatt:
Closing because I'll split into multiple PRs

PR #21, opened 2023-08-23
Fix a few bugs in the dumb encoding

oflatt:
This PR fixes 
- one bug in the structured translation (if statements do not count as scopes to break out of)
- fixes support for memory operations
- other small things

Merge after #19  and #20


src/cfg/to_structured.rs
```diff
@@ -185,7 +191,7 @@ impl<'a> StructuredCfgBuilder<'a> {
                 ContainingHistory::LoopWithLabel(label)
                 | ContainingHistory::BlockFollowedBy(label) => {
                     if label == &target {
-                        return StructuredBlock::Break(index + 1);
+                        return StructuredBlock::Break(index);
```
oflatt, line 194: Here's the bug-fix for the structured encoding

src/conversions.rs
```diff
@@ -34,6 +39,29 @@ impl Optimizer {
         }
     }
 
+    fn expr_to_argument(expr: &Expr) -> Argument {
```
oflatt, line 42: added arguments for functions, fixing a bug where we drop them

PR #23, opened 2023-08-25
add profile to nightly webpage

ajpal:
Now that https://github.com/oflatt/eggcc/pull/18 is in, wire it up to actually be used by the nightly script.
image


oflatt (approved):
Great!
Could you also add the number of instructions executed in addition to the hyperfine results? (separate PR is fine)

PR #25, opened 2023-08-26
Initial RVSDG support

ezrosent:
This PR provides initial support for conversion from bril to RVSDGs, including a restructuring pass that converts unstructured CFGs into structured ones. Missing are:

* Support for side-effects (like memory, or print)
* An egglog representation, though this PR unblocks Yihong's work on this front.
* Translation back into a CFG.

Translation back into a CFG will allow us to increase test coverage more: there are unit tests here, but writing unit tests manually is error-prone and time-consuming. Adding more translation functions will let us automate tests for larger / trickier examples.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,10 +14,13 @@ log = "0.4.19"
 thiserror = "1"
 lalrpop-util = { version = "0.19.8", features = ["lexer"] }
 petgraph = "0.6"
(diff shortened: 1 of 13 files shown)


oflatt (approved):
There are certainly some improvements we can make, but this is a great start. Thanks @ezrosent!
Merging so that we can start working on it and getting tests up and running.

PR #26, opened 2023-08-30
Expose RVSDG visualizations through command line

oflatt:
This PR allows viewing CFGs and RVSDGs through the command line with the new `debug-dir` flag. I hope to phase out most of the other command-line flags in favor of the debug information.

I also cleaned up the testing code somewhat


oflatt:
Maybe @ezrosent can review this one?

src/cfg/mod.rs
```diff
@@ -26,6 +24,22 @@ mod tests;
 pub(crate) mod structured;
 pub(crate) mod to_structured;
 
+/// Convert a program to a cfg.
+/// Loops over all the functions, translating individually.
+pub(crate) fn program_to_cfg(program: &Program) -> CfgProgram {
```
ezrosent, line 29: nit: We should decide if we want toplevel functions for converting "xs" to "ys" or if we want these to be methods on the destination type, i.e.

```rust
impl CfgProgram {
fn new(program: &Program) -> CfgProgram { ... }
}
```

I think I've added a few functions like this, but it may be more idiomatic rust to have them as methods. This comment is optional; I'd say it's up to you but we should be consistent.

src/rvsdg/mod.rs
```diff
@@ -113,7 +118,7 @@ pub(crate) enum RvsdgBody {
     },
 }
 
-pub(crate) struct RvsdgFunction {
+pub struct RvsdgFunction {
```
ezrosent, line 121: nit: Document this struct, as well as `RvsdgProgram`.

src/util.rs
```diff
@@ -63,44 +65,114 @@ impl DebugVisualizations {
...
         use std::fs::File;
         use std::io::Write;
-        for (name, content) in &[
-            ("input_cfg.dot", &self.input_cfg),
-            ("restructured_cfg.dot", &self.restructured_cfg),
-            ("rvsdg.svg", &self.rvsdg),
-        ] {
-            let mut file = File::create(format!("{}{}", prefix.as_ref(), name))?;
+
+        // make the directory, clearing it if needed
```
ezrosent, line 109: I think this is dangerous / unexpected behavior. I'd expect this functionality to just overwrite anything in the supplied directory with a colliding name, not to clear the directory of all its contents (maybe I want to just dump the files into my working directory? but then it'd remove everything in there).

oflatt: yikes you are totally right! I'll put the prefix thingy back in as well

src/cfg/mod.rs
oflatt: Good question, I'm also not sure. So far I've tried to make all the conversions simply available as methods like `Optimizer::cfg_to_rvsdg`

src/util.rs
```diff
@@ -106,11 +107,10 @@ impl DebugVisualizations {
         use std::fs::File;
         use std::io::Write;
 
-        // make the directory, clearing it if needed
-        if path.exists() {
-            std::fs::remove_dir_all(&path).unwrap();
+        // make the directory if it doesn't exist
+        if !path.exists() {
+            std::fs::create_dir_all(&path).unwrap();
```
ezrosent, line 112: we're retuning an io::Result here, can we do `std::fs::create_dir_all(&path)?;` instead?

ezrosent (approved):
One more nit.

src/util.rs
oflatt: good idea

PR #30, opened 2023-09-01
Peg visualization cleanup

Alex-Fischman:
This PR:
- Adds a PegProgram struct to mirror RVSDGs.
- Makes PegFunction::graph preserve orderings of its children (there's an ordering attribute).
- Adds pegs to the DebugVisualization struct. I think I did this right, but RVSDG viz wasn't working on the files I tested, and so PEG viz didn't work either.

### DIFF
--- a/src/peg/mod.rs
+++ b/src/peg/mod.rs
@@ -12,7 +12,7 @@
 
 pub(crate) mod simulate;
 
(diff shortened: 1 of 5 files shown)


oflatt (approved):
Great job, thanks!

PR #31, opened 2023-09-01
Profile nightly

ajpal:
Some little tweaks to #23 to make it actually work on nightlies
https://nightly.cs.washington.edu/reports/eggcc/1693597289:nightly:profile-nightly:284a8084d6/


oflatt (approved):
Thanks!

PR #32, opened 2023-09-10
Add 'state' edges to the RVSDG.

ezrosent:
This PR adds state edges to the RVSDG. This allows us to correctly handle impure function calls and `print`. It works by adding additional inputs and outputs to various constructs:

* All functions take an extra argument
* All functions return an extra output (both when defining an `RvsdgFunction` and when calling a function).
* All function calls pass the ambient "state argument" as the last argument, and reassign this argument with the value of the last output.

This allows most of the RVSDG code to proceed unmodified. With a few exceptions:

1. There was a bug in live variable analysis that, when fixed, revealed a bug in RVSDG conversion that didn't handle live variables that were still unbound. The conversion code now filters for bound variables correctly.
2. The SVG conversion code needs to handle functions with multiple output ports. I made this change.

I'm making this a draft because I'd like some guidance on how to fix the failing test here, which is related to some of the egglog roundtrip functionality that was recently added. Happy to open a proper PR once that's all fixed. 

The end result here is that even for pure functions, state arguments get plumbed through ~everywhere. The RVSDG paper describes some straightforward passes for removing these arguments where they aren't used: I think writing such a pass would be a good test for the egglog encoding of RVSDGs. I think we'll need _some_ changes but it's not clear to me if these will be minor, or more substantial.


src/peg/mod.rs
```diff
@@ -120,15 +120,19 @@ impl PegBuilder<'_> {
...
                 match &self.rvsdgs[id] {
                     // To translate a PureOp, translate all its arguments, then change ops to ids
-                    RvsdgBody::PureOp(expr) => {
+                    RvsdgBody::BasicOp(expr) => {
                         let expr = match expr {
                             Expr::Op(op, xs) => {
                                 Expr::Op(*op, xs.iter().map(|x| self.get_pegs(*x, scope)).collect())
                             }
-                            Expr::Call(f, xs) => Expr::Call(
+                            Expr::Call(f, xs, n) => Expr::Call(
```
oflatt, line 128: nit: I don't like one character names, maybe call this num_outputs

src/rvsdg/mod.rs
```diff
@@ -127,6 +144,10 @@ pub(crate) enum RvsdgBody {
 /// to nodes by their index in the vector.
 pub struct RvsdgFunction {
     /// The number of input arguments to the function.
+    ///
+    /// Functions all take `n_args + 1` arguments, where the last argument is a
+    /// "state edge" used to preserve ordering constraints to (potentially)
+    /// impure function calls.
     pub(crate) n_args: usize,
```
oflatt, line 151: Should you include the `+ 1` in n_args so that later we can have functions that don't necessarily take this argument?

oflatt (changes requested):
Looks pretty good, just had some nits and questions. 
I'm wondering if we could treat state variables as less special in the code- it seems like you have special identifiers and implicit arguments in places.

src/rvsdg/mod.rs
ezrosent: I prefer the existing structure but I'm open to changing it. I think having the state edge be initialized to `Arg(n_args)` is nicer than `Arg(n_args-1)`, because you're implicitly assuming that n_args (which is a `usize`) is at least 1.

I agree that it'd be nice to later have functions that don't necessarily take this argument (e.g. when we read them back out of egglog and we've proved that it's pure).

But in that case, I don't think it's a property of the function, it's the fact that _calls_ to the function won't have any state edges in them. In either case, we don't have to change this struct. WDYT?

src/peg/mod.rs
ezrosent: done.

src/rvsdg/from_cfg.rs
```diff
@@ -249,15 +258,21 @@ impl<'a> RvsdgBuilder<'a> {
         let mut inputs = Vec::<Operand>::new();
         let mut outputs = Vec::<Vec<Operand>>::new();
         let live_vars = self.analysis.var_state(block).unwrap();
+
+        // Not all variables that are live have necessarily been bound yet.
```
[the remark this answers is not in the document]
ezrosent: Is it the `live` used as adjective? I think that's correct here; but I rephrased anyway.

src/rvsdg/from_cfg.rs
```diff
@@ -27,20 +27,25 @@ use super::{
 
 pub(crate) fn cfg_func_to_rvsdg(cfg: &mut Cfg) -> Result<RvsdgFunction> {
     cfg.restructure();
-    let analysis = live_variables(cfg);
+    let mut analysis = live_variables(cfg);
     let dom = dominators::simple_fast(&cfg.graph, cfg.entry);
+    let state_var = analysis.intern.intern(state_id());
```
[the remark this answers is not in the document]
ezrosent: fixed.

src/peg/mod.rs
```diff
@@ -120,15 +120,19 @@ impl PegBuilder<'_> {
                 }
                 match &self.rvsdgs[id] {
                     // To translate a PureOp, translate all its arguments, then change ops to ids
```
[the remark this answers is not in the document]
ezrosent: Fixed; though I suspect we may want to go back through this PEG code to confirm. (I went ahead and renamed PegBody::PureOp)

src/cfg/mod.rs
```diff
@@ -56,7 +56,31 @@ pub(crate) enum BlockName {
...
+/// Is effectively translated to:
+///
+/// > x: int = const 1
+/// > <state> = print x <state>
+/// > y: int = const 2
+/// > <state> = print y <state>
+///
+/// `state_id` is the identifier corresponding to this <state> variable.
+pub(crate) fn state_id() -> Identifier {
+    Identifier::Num(usize::MAX)
```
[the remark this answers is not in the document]
ezrosent: Just fewer cases later on. Would you prefer I change it? I don't have a strong opinion here.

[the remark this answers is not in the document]
oflatt: Less cases is good!

src/rvsdg/mod.rs
oflatt: That sounds like a good argument but I'm a little confused still.
Does that mean that we can call a function without the state argument, but the function still takes that argument implicitly? seems like an arity mismatch

oflatt (approved):
Great job, merging!

src/rvsdg/mod.rs
oflatt: @ezrosent I'm still unclear on this

ezrosent: Sorry; I didn't revisit this page after merging (in the future, I'd be happy to try and resolve these threads before merging the PR).

> Does that mean that we can call a function without the state argument, but the function still takes that argument implicitly? seems like an arity mismatch

Right now, all functions will take a state argument, so whether or not it counts in `n_args` or not is just a matter of convention.

If we remove state arguments as part of optimizations in egglog, then there won't be any mention of `Arg(n_args)` in the body and we will have to make sure not to extract any calls to this function that pass a state edge by mistake.

Those are all real concerns, but I don't see them as being harder or easier with `n_args` being what it is here. Maybe that'll become clearer once we extract more things.

I don't feel strongly about this though; I'm fine changing it.

oflatt: Oh, I see
Yeah I see what you are saying, I think it doesn't matter except that it's confusing that you have the same `n_args` whether you are passing the state edge or not.

PR #33, opened 2023-09-14
Unify testing infra, debug visualizations, and command line arguments

oflatt:
This PR refactors all three of the testing infra, debug visualizations, and command line arguments to use the same code path. This cleans things up significantly and makes sure we are snapshot testing the debug visualizations.

Our next step should be to carefully look at these snapshots (I already know the two_fns snapshot is missing a whole function in the viz). Then we should move more tests out of the `failing` folder as we fix bugs


src/lib.rs
```diff
@@ -59,6 +60,32 @@ fn run_command_with_stdin(command: &mut std::process::Command, input: String) ->
     .unwrap()
     .to_string()
 }
+#[derive(Clone, Copy)]
+pub enum RunType {
+    StructuredConversion,
+    RvsdgConversion,
+    NaiiveOptimization,
```
ezrosent, line 67: sp: https://en.wiktionary.org/wiki/naive ?

src/util.rs
```diff
@@ -39,143 +43,216 @@ pub(crate) fn parse_from_string(input: &str) -> bril_rs::Program {
...
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .spawn()?;
+
+    let stdin = child.stdin.as_mut().expect("Failed to open stdin");
+    write!(stdin, "{}", input)?;
+
+    let output = child.wait_with_output()?;
+    match output.status.code() {
+        Some(0) => Ok(String::from_utf8(output.stdout).map_err(|e| {
```
ezrosent, line 107: All these ErrorKind::Others point to it being better for us to have our own error type here, and then return an `anyhow::Result` to capture the io errors as well.

src/util.rs
```diff
@@ -39,143 +43,216 @@ pub(crate) fn parse_from_string(input: &str) -> bril_rs::Program {
...
+        match s {
+            "structured" => Ok(RunType::StructuredConversion),
+            "rvsdg" => Ok(RunType::RvsdgConversion),
+            "naiive" => Ok(RunType::NaiiveOptimization),
+            _ => Err(format!("Unknown run type: {}", s)),
+        }
+    }
 }
 
-impl DebugVisualizations {
```
ezrosent, line 62: It's not clear to me how easy it is to add ad-hoc calls to this code now. One thing I use DebugVisualizations for is just adding a call to it in a test that I'm debugging. Can we still do that here?

ezrosent (changes requested):
Minor comments, largely optional

ezrosent:
Why do we want to snapshot test the debug visualizations?

src/lib.rs
oflatt: hmm idk, I feel like it's spelled Naiive

src/util.rs
oflatt: I was thinking about that
Would it be better for us to wrap it up in an `EggCCError`? Or we could convert all of our code to use anyhow, which would remove the need for `EggCCError`

src/util.rs
oflatt: I'll change the `visualize` function to take a bril program instead of a file path. Does that recover the functionality?

oflatt:
I don't think we necessarily want to snapshot test all the visualizations in the future- but I do want to snapshot test the rvsdg svg and structured conversion.
Before I had a separate code path that did this but now it's all the same. Also we should only run snapshot tests on the files in the `small` folder, let me put that check back in.

src/util.rs
oflatt: I've refactored the code so that you can either provide a bril file or the actual bril code (along with the name for the code and the arguments)

src/util.rs
oflatt: let's fix this one another time

ezrosent (approved):

PR #35, opened 2023-09-20
Add types to the rvsdg and egglog encoding

oflatt:


src/rvsdg/mod.rs
```diff
@@ -92,17 +92,17 @@ pub(crate) type Id = usize;
 #[derive(Debug, PartialEq)]
 pub(crate) enum Expr<Op> {
```
ezrosent, line 93: optional: should we name the fields now that there are so many (similar to the various `Ops` in the bril Rust bindings?)

src/rvsdg/from_cfg.rs
```diff
@@ -75,12 +79,15 @@ pub(crate) fn cfg_func_to_rvsdg(cfg: &mut Cfg) -> Result<RvsdgFunction> {
     })
 }
 
+pub(crate) type FunctionTypes = HashMap<String, Option<Type>>;
```
ezrosent, line 82: Why Option<Type> (and document it here)

src/rvsdg/mod.rs
```diff
@@ -209,28 +214,38 @@ pub enum EgglogFunctionResult {
...
-        };
-        fn from_ty(ty: &Type) -> egglog::ast::Expr {
-            match ty {
-                Type::Int => Call("IntT".into(), vec![]),
-                Type::Bool => Call("BoolT".into(), vec![]),
-                Type::Float => Call("FloatT".into(), vec![]),
-                Type::Char => Call("CharT".into(), vec![]),
-                Type::Pointer(ty) => Call("PointerT".into(), vec![from_ty(ty.as_ref())]),
+        let f = |operands: &Vec<Operand>, ty: Option<Type>| {
+            let mut res = vec![];
```
ezrosent, line 231: nit: `Vec::with_capacity(operands.len() + ty.is_some() as usize)` (or similar)

ezrosent (approved):
Let me know if you want another round of feedback; but I'll leave it up to you if you want to merge after responding.

src/rvsdg/mod.rs
oflatt: We probably should (I wish there was a cool vscode refactor for this)

src/rvsdg/from_cfg.rs
oflatt: Left a comment- functions don't always return a value so for these we use `None`. Bril doesn't have a void type

PR #38, opened 2023-09-26
Make PEGs work with print edges and remove naive tests

Alex-Fischman:
- Create a new `RunType` variant for PEGs.
- Fix the testing to run the optimized bril if it's supported, instead of always running the input program.
- Remove the naive tests and add PEG tests.
- Fix a `peg2dot` shadowing bug.
- Add a state edge to `PegFunction`, fixing the tests.


Alex-Fischman:
Since the review request, I have added:
- Function names to `PegFunction` and `RvsdgFunction`
- Simulation of full `PegPrograms`
- Simulation of PEG function calls
- Simulation of PEG print statements
- Simulation of PEG `div` instructions

I have also:
- Fixed syntax errors in the `.dot` output
- Added an automatic call to the `dot` command line utility
- Added installation of the `graphviz` library to CI for `dot`

oflatt (approved):
Great job, we will have to figure out why it's slow next

PR #39, opened 2023-09-28
bump egglog revision

ezrosent:
Larger diff because this changed the peg visualizations. Spot-checks look reasonable.


oflatt (approved):
good!

PR #40, opened 2023-09-28
Convert cfgs to Bril

oflatt:
This PR adds a conversion from our CFG representation back to bril. It also distinguishes between `Simple` CFGs, which only branch on bril booleans, and `Switch` cfgs, which are used internally to translation to RVSDG.

This distinguishment is only enforced by asserts, but there's also a type alias for the two to keep them apart.

Finally, I changed the peg snapshots to use dot files for stability, because different versions of graphviz generate different svgs.


src/cfg/mod.rs
```diff
@@ -24,25 +26,59 @@ pub(crate) type NodeSet = <StableDiGraph<BasicBlock, Branch> as Visitable>::Map;
...
+impl SimpleCfgFunction {
+    pub fn convert_to_switch(self) -> SwitchCfgFunction {
+        SwitchCfgFunction {
+            args: self.args,
+            graph: self.graph,
+            entry: self.entry,
+            exit: self.exit,
+            name: self.name,
+            return_ty: self.return_ty,
+            phantom: PhantomData,
```
oflatt, line 76: Needs this PhantomData thingy to keep the Simple apart from the Switch. I'm not sure I love this, but it's kind of a cool trick. This way we get to share a lot of code, and we just don't use annotations for the simple kind.

src/cfg/mod.rs
```diff
@@ -261,11 +307,80 @@ impl Cfg {
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum SimpleBranch {
```
oflatt, line 311: We can also always get a simple branch from these simpler cfgs

tests/files.rs
```diff
@@ -1,18 +1,30 @@
...
+
+use eggcc::util::{Run, RunType, TestProgram};
 use insta::assert_snapshot;
 use libtest_mimic::Trial;
 
 fn generate_tests(glob: &str) -> Vec<Trial> {
     let mut trials = vec![];
+
     let mut mk_trial = |run: Run, snapshot: bool| {
+        let snapshot_configurations: HashSet<RunType> = vec![
```
oflatt, line 11: Only some of the configurations need snapshots

src/cfg/to_bril.rs
```diff
@@ -0,0 +1,129 @@
...
+                        label: format!("{}", self.graph[node].name),
+                        pos: None,
+                    });
+                    // rest of the block
+                    self.node_to_bril(node, &mut func);
+                }
+            });
+
+        // now do the exit at the end
+        func.instrs.push(Code::Label {
```
ezrosent, line 55: nit: how about we factor this out into a function `get_label` that takes the graph + node and outputs the `Code` ?

src/cfg/to_bril.rs
```diff
@@ -0,0 +1,129 @@
...
+
+        // The order of this traversal does not matter, just need to loop over the blocks
+        DfsPostOrder::new(&self.graph, self.entry)
+            .iter(&self.graph)
+            .for_each(|node| {
+                // don't do the exit or entry
+                if node != self.entry && node != self.exit {
+                    // Add a label for the block
+                    func.instrs.push(Code::Label {
+                        label: format!("{}", self.graph[node].name),
```
ezrosent, line 46: nit: I think it'd be easier to read (and refactor) if we had a separate `label_name`  function that converted an identifier to a label. We use this convention of `format!("{}"` a few different places in this file, and it could be difficult to clean up in the future if we needed to change the formatting for some reason.

src/cfg/to_bril.rs
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+    }
+}
+
+impl BasicBlock {
+    // Converts this block to bril, not including jumps to other block
+    // or edges.
+    pub fn to_bril(&self) -> Vec<Code> {
+        let mut res = vec![];
```
ezrosent, line 103: nit, optional: `let mut res: Vec<Code> = self.instrs.iter().cloned().map(Code::Instruction).collect();`

Allows the vector to be pre-sized correctly.

tests/files.rs
ezrosent: I don't think you need the `vec!`; I think you can just iterate over the array.

ezrosent (commented):
Mostly nits. One more substantial comment on the phantom types.

src/cfg/mod.rs
```diff
@@ -24,25 +26,59 @@ pub(crate) type NodeSet = <StableDiGraph<BasicBlock, Branch> as Visitable>::Map;
...
-pub struct CfgProgram {
-    pub functions: Vec<Cfg>,
+pub struct CfgProgram<CfgType> {
+    pub functions: Vec<CfgFunction<CfgType>>,
+}
+
+/// Simple programs only branch on booleans defined in bril
+pub type SimpleCfgProgram = CfgProgram<Simple>;
+/// Switch programs can also branch on values defined in annotations
+pub type SwitchCfgProgram = CfgProgram<Switch>;
```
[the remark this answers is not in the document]
ezrosent: Ah, I guess "Ret" is still in the simple CFG.

src/cfg/mod.rs
oflatt: Yeah, that's right. Thanks for the example code, I was considering something similar and I like the guarantee that we have certain annotations. 

I think for now let's leave it as is. I think there are several differences I'd like to enforce between the two but converting between them is cumbersome. Can revisit this question in the future

src/cfg/mod.rs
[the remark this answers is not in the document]
oflatt: Yeah, it's kind of confusing that we still have `Ret` but it's convenient to keep them at the footer.

src/cfg/mod.rs
```diff
@@ -261,11 +307,80 @@ impl Cfg {
...
+        arg: Identifier,
+        then_branch: BlockName,
+        else_branch: BlockName,
+    },
+}
+
+impl SimpleCfgFunction {
+    pub fn get_branch(&self, node: NodeIndex) -> SimpleBranch {
+        let outgoing = self.graph.edges(node);
+        match &outgoing.collect::<Vec<_>>().as_slice() {
```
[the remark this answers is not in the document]
oflatt: Had weird type errors, we can fix later

src/cfg/to_bril.rs
oflatt: Didn't do this one

oflatt:
Thanks for review @ezrosent I addressed all your nits that I marked as resolved (:

ezrosent (approved):

PR #41, opened 2023-09-28
Add RVSDG subst functions

rtjoa:
## Summary

Adds the following functions, for use in writing passes over RVSDG
```
;;                   e  [ x -> v ]
(function SubstExpr (Expr i64 Operand) Expr)
(function SubstOperand (Operand i64 Operand) Operand)
(function SubstBody (Body i64 Operand) Body)
(function SubstVecOperand (VecOperand i64 Operand) VecOperand)
(function SubstVecVecOperand (VecVecOperand i64 Operand) VecVecOperand)
```

Updated the egglog revision we depend on as this feature relies on `vec-set`: https://github.com/egraphs-good/egglog/pull/239

## Testing

Added tests to [`src/rvsdg/tests.rs`](https://github.com/rtjoa/eggcc/blob/subst/src/rvsdg/tests.rs)

### DIFF
--- a/src/rvsdg/schema.egg
+++ b/src/rvsdg/schema.egg
@@ -3,10 +3,11 @@
 (datatype Operand)
 (datatype Body)
 
-(sort VecOperand (Vec Operand))
-(datatype VecOperandWrapper
-  (VO VecOperand))
-(sort VecVecOperand (Vec VecOperandWrapper))
+(sort VecOperandBase (Vec Operand))
+(datatype VecOperand (VO VecOperandBase))
+
+(sort VecVecOperandBase (Vec VecOperand))
+(datatype VecVecOperand (VVO VecVecOperandBase))
 
 ;; Type
 (datatype Type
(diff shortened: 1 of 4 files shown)


src/rvsdg/schema.egg
```diff
@@ -48,6 +49,97 @@
 (function Gamma (Operand VecOperand VecVecOperand) Body) ;; branching
 (function Theta (Operand VecOperand VecOperand) Body) ;; loop
 
+;; Substitution
+
+;;                   e  [ x -> v ]
+(function SubstExpr (Expr i64 Operand) Expr)
```
oflatt, line 55: I don't see why this shouldn't be
`(function SubstExpr (Expr Operand Operand) Expr)`
This would allow you to substitute operands for other operands, not just arguments.

src/rvsdg/schema.egg
```diff
@@ -48,6 +49,97 @@
...
+  ))
+
+(rule
+  (
+    (= f (SubstVecVecOperandHelper (VVO vec) x v i))
+    (= i (vec-length vec))
+  )
+  ( 
+    (union (SubstVecVecOperandHelper (VVO vec) x v i) (VVO vec))
+  ))
```
oflatt, line 143: Can you add a ruleset for the subst rules to make it more usable?
I'd also like the existing egglog encoding to run this ruleset to saturation in-between runs of the normal ruleset.

src/rvsdg/schema.egg
```diff
@@ -48,6 +49,97 @@
...
+
+(rewrite (SubstBody (PureOp e) x v) (PureOp (SubstExpr e x v)))
+;; Subst doesn't cross regions - so we subst into the inputs but not outputs
+;; Node that a Gamma node's idx is on the outside, so it gets affected, but not
+;; a Theta node's predicate
+(rewrite (SubstBody (Gamma pred inputs outputs) x v) (Gamma (SubstOperand pred x v) (SubstVecOperand inputs x v) outputs))
+(rewrite (SubstBody (Theta pred inputs outputs) x v) (Theta pred (SubstVecOperand inputs x v) outputs))
+
+;; params: vec, var, op, next index to subst
+;; rtjoa: TODO: implement by mapping internally so they're not O(n^2) time
```
oflatt, line 90: Why is it n^2 time?

oflatt (changes requested):
Requested some changes, this is very close to what we want!

src/rvsdg/schema.egg
rtjoa: Because `vec-set` is linear, and we set indices one-at-a-time (this is what I mentioned in person, about how we should eventually either use a vec & built-in map, or a cons-list & map rules, but not mix)

src/rvsdg/schema.egg
rtjoa: Some initial rambling, but let's talk about this in person:

I think if we allow subst target arbitrary e-classes, rather than concrete args, then these rules become less "pure computation" and it becomes harder to "delete subst once we know the target doesn't occur," since something might get unioned in the future.

src/rvsdg/schema.egg
rtjoa: Added all rules to the `subst` ruleset, added the existing rules to constant-fold +/* to the `cfold` ruleset, and made the schedule the following:
```rs
let schedule = format!("(run-schedule (repeat {} cfold subst))", self.num_iters);
```

src/rvsdg/schema.egg
oflatt: Okay, semantics too hard lets just do args for now

src/rvsdg/schema.egg
oflatt: Okay I understand thanks

oflatt (approved):
good job!

PR #42, opened 2023-09-29
PEG Interpreter Memoization

Alex-Fischman:
This PR cleans up and then speeds up the PEG interpreter. It is now fast enough to run the fib-shape test with n=10 as part of the test suite. It's still honestly not that fast, but I don't think it needs to be faster than it is now for the foreseeable future.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -18,11 +18,6 @@ struct Args {
     #[clap(long)]
     interp: bool,
 
-    /// Translate the input program to a PEG and run it.
-    /// Throws an error if used with the interp flag.
-    #[clap(long)]
-    peg_interp: bool,
-
     /// Path that eggcc will put interp profile results
     #[clap(long)]
     profile_out: Option<PathBuf>,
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -18,6 +18,11 @@ struct Args {
     #[clap(long)]
     interp: bool,
 
+    /// Translate the input program to a PEG and run it.
+    /// Throws an error if used with the interp flag.
+    #[clap(long)]
+    peg_interp: bool,
```
oflatt, line 24: We don't need this command line flag because we already have the `run-mode` and `interp` flags, which should already just work!

src/peg/simulate.rs
```diff
@@ -1,70 +1,94 @@
...
+    args: Vec<Literal>,
+    indices: Rc<Indices>,
+    func: usize,
+    program: &'a PegProgram,
+    stdout: Rc<RefCell<String>>,
+    memoizer: Rc<RefCell<Memoizer>>,
 }
 
-impl PegBody {
+type Memoizer = HashMap<(usize, usize, Rc<Indices>), Literal>;
```
oflatt, line 68: what are these usize arguments? I think this and Indices could use some documentation

src/peg/simulate.rs
```diff
@@ -1,70 +1,94 @@
...
-    fn simulate(
-        &self,
-        args: &[Literal],
-        nodes: &[PegBody],
-        indices: &Indices,
-        program: &PegProgram,
-        stdout: &mut String,
-    ) -> Option<Literal> {
-        match self {
+    fn simulate_body(&self, i: usize) -> Option<Literal> {
```
oflatt, line 80: Good change with putting stuff in a simulator. what is `i`?
Nit: I don't like single-letter variables because they are hard to search for and error-prone

oflatt (changes requested):
Had some small nits, overall looks great.
Have you tried bumping the fib iteration number to something large?

src/main.rs
Alex-Fischman: `interp` means "run the optimized Bril", `peg-interp` means "run the unoptimized PEG". I think that it's worth separating the flags because of the optimized/unoptimized distinction, not the Bril/PEG distinction.

oflatt: Actually, I usually think of `interp` as "run the result". The `RunType` determines if it is optimized or not optimized

oflatt (dismissed):
Still think that we should get rid of peg_interp

oflatt (approved):
Thanks!

PR #43, opened 2023-09-29
Supporting functions in egglog's encoding of RVSDG

yihozhang:


oflatt (changes requested):
Did you add the return type and argument types to the `RvsdgFunction` struct? I think I'll need those

oflatt (approved):
Thank you! This is great!

PR #46, opened 2023-10-04
Egglog RVSDG Roundtrip

oflatt:
Merge after #44


Cargo.toml
```diff
@@ -9,7 +9,7 @@ name = "files"
 
 
 [dependencies]
-egglog = { git = "https://github.com/egraphs-good/egglog", rev = "cec39af" }
+egglog = { git = "https://github.com/hatoo/egglog/", rev = "c83a509" }
```
oflatt, line 12: Don't merge until this makes it to egglog!

src/peg/mod.rs
```diff
@@ -23,7 +20,7 @@ mod tests;
 #[derive(Debug, PartialEq)]
 pub(crate) enum PegBody {
     /// A basic expression.
-    BasicOp(Expr<Id>),
+    BasicOp(RvsdgExpr<Id>),
```
ezrosent, line 23: We're still using this for PEGs, right? should we name this something else?

Also, to Anjali's point earlier, this sort of rename is really easy to review if it's the entirety of the PR (w/the remaining changes stacked on top).

src/rvsdg/from_egglog.rs
```diff
@@ -0,0 +1,231 @@
...
+            }
+        } else {
+            panic!("expect an operand, got {op}")
+        }
+    }
+
+    fn expr_to_vec_operand(vec: &Expr, bodies: &mut Vec<RvsdgBody>) -> Vec<Operand> {
+        let Expr::Call(func, args) = vec else {
+        panic!("Expected a VO, got {vec}")
+      };
```
ezrosent, line 28: formatting looks off here.

src/rvsdg/from_egglog.rs
```diff
@@ -0,0 +1,231 @@
...
+            }
+        } else {
+            panic!("expect a function, got {expr}")
+        }
+    }
+}
+
+fn vec_map<T>(inputs: &Expr, mut f: impl FnMut(&Expr) -> T) -> Vec<T> {
+    use Expr::*;
+    let mut inputs: &Expr = inputs;
```
ezrosent, line 206: declare the inputs parameter as `mut` instead?

src/rvsdg/from_egglog.rs
```diff
@@ -0,0 +1,231 @@
...
+                }
+                _ => panic!("expect a function, got {expr}"),
+            }
+        } else {
+            panic!("expect a function, got {expr}")
+        }
+    }
+}
+
+fn vec_map<T>(inputs: &Expr, mut f: impl FnMut(&Expr) -> T) -> Vec<T> {
```
ezrosent, line 204: I think this function should be documented. I don't think it's obvious from the signature the way in which it is a map operation.  I'm not positive about what it does, My guess is:

> Call `f` on each element of `inputs`, where `inputs` must be a fully expanded vector egglog vector expression.

src/rvsdg/from_egglog.rs
```diff
@@ -0,0 +1,231 @@
...
+                        Some(Self::egglog_expr_to_ty(ty)),
+                    )
+                }
+                ("Const", [ty, _const_op, lit]) => RvsdgExpr::Const(
+                    // todo remove the const op from the encoding because it is always ConstOps::Const
+                    ConstOps::Const,
+                    Self::egglog_expr_to_literal(lit),
+                    Self::egglog_expr_to_ty(ty),
+                ),
+                ("PRINT", [opr1, opr2]) => {
```
ezrosent, line 99: I think bril supports more than one arg to print (state edge or no), see https://github.com/oflatt/eggcc/blob/main/tests/brils/failing/char/char.bril

src/rvsdg/live_variables.rs
```diff
@@ -173,6 +173,7 @@ impl VarSet {
         self.vars.count_ones(..)
     }
 
+    #[allow(dead_code)]
```
ezrosent, line 176: If these are dead, can we remove them? (if they're just used in tests, let's annotate them as `#[cfg(test)]`

src/rvsdg/schema.egg
```diff
@@ -41,6 +41,7 @@
 (function not (Type Operand Operand) Expr)
 (function and (Type Operand Operand) Expr)
 (function or (Type Operand Operand) Expr)
+(function PRINT (Operand Operand) Expr)
```
ezrosent, line 44: ah, so print isn't taking a vec here?

tests/snapshots/files__fib_shape-rvsdg-optimize.snap
```diff
@@ -0,0 +1,220 @@
...
+  v1: int = const 0;
+  v2: int = const 1;
+  v3: int = id v1;
+  v4: int = id v2;
+  v5: int = id v0;
+  jmp .__7__;
+.__130__:
+  v128: int = const 1;
+  v129: bool = eq v121 v128;
+  v68: int = id v87;
```
ezrosent, line 15: (no change needed): this is what they mean by "copy insertion" in the "Perfect Reconstruction" paper. 

Sounds like there are standard algorithms here to clean this up.

ezrosent (changes requested):
Seems like this is mostly renames / moves?

tests/snapshots/files__fib_shape-rvsdg-optimize.snap
oflatt: ohhhh makes more sense to me now

oflatt:
Yep, this is mostly a cleanup PR

Cargo.toml
oflatt: fixed

src/peg/mod.rs
oflatt: Oh I didn't realize it was used in PEGs.
Yeah, renaming should be separate I agree.
I'll do another PR for that!

src/rvsdg/from_egglog.rs
oflatt: good idea, that's right

src/rvsdg/from_egglog.rs
oflatt: yeah we just support one for now, we can fix in the future

src/rvsdg/live_variables.rs
oflatt: yep, done

oflatt:
Addressed your comments @ezrosent, another rename PR incoming

ezrosent (approved):

PR #49, opened 2023-10-06
Refactor substitution rules into their own file

oflatt:
I wish we had egglog modules, but for now I think separating into files will have to do.


yihozhang (approved):
LGTM!

PR #50, opened 2023-10-06
Add basic constant folding rules to eggcc

oflatt:


src/rvsdg/egglog_optimizer/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+    "(run 3)".to_string()
+}
+
+#[derive(Debug, PartialEq, Clone)]
+struct BrilOp {
+    op: &'static str,
+    egglog_op: &'static str,
+    // so far we don't need more than 2 types
+    // if the field is None, it is beyond the number of inputs
+    input_types: [Option<Type>; 2],
```
yihozhang, line 26: this looks a bit uncommon to me. Why not `Vec<Type>`?

src/rvsdg/egglog_optimizer/constant_fold.rs
```diff
@@ -0,0 +1,20 @@
...
+
+pub(crate) fn constant_fold_egglog() -> String {
+    let mut res = vec![];
+
+    for bril_op in BRIL_OPS {
+        let op = bril_op.op;
+        let egglog_op = bril_op.egglog_op;
+        if bril_op.num_inputs() == 2 {
+            res.push(format!(
+                "(rewrite ({op} ty
```
yihozhang, line 11: Alas if only we had metaprogramming in egglog.

yihozhang (approved):

src/rvsdg/egglog_optimizer/mod.rs
oflatt: I'm declaring a global type so it needs to have a fixed size apparently

src/rvsdg/egglog_optimizer/constant_fold.rs
oflatt: alas

PR #51, opened 2023-10-06
Fixing ruleset definitions in naive mode

yihozhang:
Thanks @clyben for spotting this.


oflatt:
Ripping it out completely in #52

PR #52, opened 2023-10-06
Rip out naiive optimizer from eggcc

oflatt:
This PR rips out the "naiive" optimizer from eggcc, which is not very good and also buggy.
Unfortunately, this means that going forward there will be fewer opportunities to spell naiive with two i's.


ajpal (approved):

src/lib.rs
```diff
@@ -137,9 +131,7 @@ impl Optimizer {
         let prog = Program::try_from(abstract_prog)
             .map_err(|err| EggCCError::ConversionError(err.to_string()))?;
 
-        // HACK: Check for uninitialized variables by looking for `__undefined`
-        // variables in the program.
-        Optimizer::check_for_uninitialized_vars(&prog)?;
+        check_for_uninitialized_vars(&prog)?;
```
ajpal, line 134: is this not a hack anymore?

oflatt: It's a good safety check

PR #54, opened 2023-10-09
add annotations to debug output for CFG

ezrosent:
There's custom debug output for basic blocks that reconstructs the bril body for the main instructions. This makes the debug output much more readable, but it is missing the content of annotations. This PR adds the annotations back.


oflatt (approved):
Good catch, we should definitely be showing these.

Issue #57, opened 2023-10-09
RVSDG optimizer fails to share sub-loops


oflatt:
Here's a benchmark that results in poor output:
```@main() {
i: int = const 0;
one: int = const 1;
three: int = const 3;
.loop:
  i: int = add i one;
  cond: bool = lt i three;
  br cond .loop .done;
.done:
  print i;
}
```

Looking at the optimized RVSDG, we can see that we get two loops instead of just one. My guess is that we are turning an `Expr` into an RVSDG instead of using the `Term` and `TermDAG` that egglog gives us to share sub-terms. I think @yihozhang wrote that code, so I'm giving you this one.

yihozhang:
Can you confirm this is fixed by your PR #70 ? @oflatt

oflatt:
I can confirm, this test is called `unroll_and_constant_fold.bril` in main

PR #58, opened 2023-10-09
Add benchmark from #57

oflatt:
This adds the benchmark from #57 so that we can fix it and track progress. Will also be a good loop unrolling benchmark.


yihozhang (approved):
LGTM!

PR #61, opened 2023-10-11
Fix eager PEG phi node eval

### LINKED ISSUE #59: Failing peg benchmark
The following program should print `1\n` but it prints `1\n2\n` in the peg interpreter
```
@main {
  a: int = const 1;
  b: int = const 2;
  condition: bool = lt a b;
  br condition .then .else;
.then:
  thenval: int = const 1;
  print thenval;
  jmp .end;
.else:
  elseval: int = const 2;
  print elseval;
  jmp .end;
.end:
}
```

Alex-Fischman:
Fixes #59.

Phi nodes were being evaluated eagerly, now they're not.

### DIFF
--- a/src/peg/simulate.rs
+++ b/src/peg/simulate.rs
@@ -144,13 +144,10 @@ impl Simulator<'_> {
                 }
             }
             PegBody::Phi(c, x, y) => {
-                let c = self.simulate_body(*c);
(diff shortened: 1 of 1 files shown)


oflatt (approved):
Good fix!

PR #62, opened 2023-10-11
Add RVSDG shift rules

rtjoa:
## Summary

Adds the following functions, for use in writing passes over RVSDG.
```
;; Within e, replace all (Args x) where x > last-unshifted
;; with (Args (x + amt))
;;                             e             last-unshifted  amt
(function ShiftExpr           (Expr          i64             i64) Expr)
(function ShiftOperand        (Operand       i64             i64) Operand)
(function ShiftBody           (Body          i64             i64) Body)
(function ShiftVecOperand     (VecOperand    i64             i64) VecOperand)
(function ShiftVecVecOperand  (VecVecOperand i64             i64) VecVecOperand)```
```

The rules that execute them are placed in ruleset `shift`.

## Testing

Added tests to [`src/rvsdg/tests.rs`](https://github.com/rtjoa/eggcc/blob/subst/src/rvsdg/tests.rs)

### DIFF
--- a/src/rvsdg/egglog_optimizer/mod.rs
+++ b/src/rvsdg/egglog_optimizer/mod.rs
@@ -8,6 +8,7 @@ pub fn rvsdg_egglog_code() -> String {
     let code = vec![
         include_str!("schema.egg").to_string(),
         include_str!("subst.egg").to_string(),
+        include_str!("shift.egg").to_string(),
         constant_fold_egglog(),
     ];
     code.join("\n")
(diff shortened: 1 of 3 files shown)


oflatt (changes requested):
This looks good, but I think it would be worth doing the templated rust string thingy before we merge. Make a module called `shift.rs`

Also please put line-breaks in your rules. The right-hand side of a rule should be on the next line at least.

oflatt (approved):
Great! We discussed doing rust for these, @rtjoa promised to do it in another PR

PR #63, opened 2023-10-11
Fix RVSDG->CFG translation

oflatt:
This PR completely rewrites RVSDG->CFG translation to a new, much nicer algorithm. The new algorithm returns a `TranslationResult`, representing a self-contained CFG with a start and end. The new algorithm is much easier to reason about.

Unfortunately, it adds a lot of unnecessary blocks because any node has to make new blocks. So we'll need an optimization pass in the future to fix those. Opening an issue

Also in the PR is a small fix to restructuring from @ezrosent, thank you!


ezrosent:
High-level comment:

* This PR seems to "glue" two blocks together by doing a jump. That creates a lot of jumps even when there's no control flow in the original program.
* Is there a reason that we can't just concatenate the two basic blocks together when we merge them?

oflatt:
Good points
I want to do both of those things as a separate pass after the fact. Yes, we could fix both those problems, but there would still be more cases of direct jumps to handle. See #66

src/rvsdg/to_cfg.rs
```diff
@@ -27,54 +27,31 @@ use crate::{
...
 /// Represents the result of a RVSDG computation
 #[derive(Clone, Debug)]
 enum RvsdgValue {
     StateEdge,
     BrilValue(String, Type),
 }
 
+/// The return type of translating a part of an RVSDG
+/// Translating an RVSDG node always creates new blocks.
+/// These may be redundant or have a single input and output, but these
```
ezrosent, line 39: Turn this into a TODO?

src/rvsdg/to_cfg.rs
```diff
@@ -27,54 +27,31 @@ use crate::{
...
+/// are removed in an optimization pass.
+#[derive(Clone, Debug)]
+struct TranslationResult {
+    start: NodeIndex,
+    end: NodeIndex,
+    values: Vec<RvsdgValue>,
```
ezrosent, line 45: I think it'd help to include an intuitive translation of what these fields mean, particularly `values`.

In particular, I think the RVSDGs that we currently generate have lots of variables plumbed through that are "redundant" for a given block; if we start hoisting those inputs out, is our use of `values` in this file still correct?

src/rvsdg/to_cfg.rs
```diff
@@ -154,84 +131,157 @@ impl RvsdgFunction {
...
+                res2.start,
+                Branch {
+                    op: BranchOp::Jmp,
+                    pos: None,
+                },
+            );
+        }
+        TranslationResult {
+            start: results[0].start,
+            end: results[results.len() - 1].end,
```
ezrosent, line 201: nit, optional: I think doing `results.last().unwrap()` is a bit easier to read than `results[results.len()-1]`.

src/rvsdg/to_cfg.rs
```diff
@@ -154,84 +131,157 @@ impl RvsdgFunction {
...
+        }
+    }
+
+    /// Like [`SequenceResults`], but combines all of the results into
+    /// a multi-output [`TranslationResult`].
+    fn combine_results(&mut self, results: &[TranslationResult]) -> TranslationResult {
+        // first sequence results
+        let sequenced = self.sequence_results(results);
+        // now make a vec of all the values
+        let mut values = vec![];
```
ezrosent, line 212: `Vec::with_capacity(results.len()) ?`

ezrosent (changes requested):

src/rvsdg/to_cfg.rs
oflatt: done

src/rvsdg/to_cfg.rs
oflatt: Good idea
I added a comment. The idea is that `values` is a list of bindings for the outputs of the node. For single outputs, this is a vec of length one.
To answer your question, I don't see why that would be a problem. If an input is hoisted out, it's still a valid RVSDG and the node will have one less output.

src/rvsdg/to_cfg.rs
oflatt: changed to an iterator instead

ezrosent (approved):
sounds good! (though please fix the test failures)

oflatt:
This is strange, I can't replicate the failure locally

oflatt:
Nvm needed to merge main

oflatt:
nevermind just needed to merge main

PR #70, opened 2023-10-12
Share subexpressions in translating back from RVSDG

oflatt:
This PR changes the code to produce much cleaner output by actually sharing subexpressions when translating from egglog to RVSDGs

Most of this PR is changing `Expr` to `Term`. Finally it slaps on a cache.


yihozhang (approved):
LGTM!

PR #72, opened 2023-10-12
Fix up bug in arities

oflatt:
This PR fixes a couple bugs. First, in RVSDG->CFG it fixes an arity mismatch. Second, it fixes PRINT visualization in RVSDGS to only have one output port as they should.

I added a reproduction test


Alex-Fischman (approved):
Looks good to me.

Issue #71, opened 2023-10-12
Failure in CFG -> RVSDG


kirstenmg:
Following Bril causes error in `to_cfg.rs`:

```
# ARGS: 2
@main(n: int) {
    curr: int = const 1;
    prev: int = const 1;
    zero: int = const 0;
    i: int = const 1;

    print prev;

    is_zero: bool = eq n zero;
    br is_zero .for.done .fencepost;

.fencepost:
    print curr;

.for.cond:
    less: bool = lt i n;
    br less .for.body .for.done;

.for.body:
    tmp: int = add curr prev;
    print tmp;
    prev: int = id curr;
    curr: int = id tmp;

    one: int = const 1;
    i: int = add i one;
    jmp .for.cond;

.for.done:

}
```

Error:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `NodeIndex(4)`,
 right: `NodeIndex(2)`', src/rvsdg/from_cfg.rs:327:17
```

kirstenmg:
Is the same bug as #69 ?

ezrosent:
Thanks for the example! 

I think #69 is about converting RVSDG => CFG, but this error happens in the other direction so I suspect it's a separate bug. I'll try and get to #65 in the next couple days and will try this example too and report back with my findings.

oflatt:
Did this one still error?

PR #75, opened 2023-10-13
Add brils, mostly failing

kirstenmg:
This PR adds more bril programs as examples, including some that demonstrate current limitations/bugs.

`cycle.bril` fails because we always run the interpreter.
`fib_recursive.bril` fails because functions aren't supported yet.
`fib.bril` fails, see #71 
`duplicate_branch.bril` succeeds.


oflatt:
Looks like you need to run `cargo insta review`
You should look at the results and check that they look reasonable for each kind of output- it's okay if they don't make sense you can accept them.

oflatt (approved):
Great!
Now we have 3 implementations of fib, we can rename them in the future.

PR #77, opened 2023-10-13
Cleanup indirection in blocks introduced by RVSDG->CFG translation

### LINKED ISSUE #66: RVSDG->CFG translation introduces many blocks
To make the translation clean, #63 introduces many blocks with a single input edge and single output edge. We should collapse these with an optimization for clean output.

oflatt:
Closes #66


src/rvsdg/optimize_direct_jumps.rs
```diff
@@ -0,0 +1,113 @@
...
+};
+
+use crate::cfg::{BasicBlock, Branch, Simple, SimpleCfgFunction, SimpleCfgProgram};
+
+struct JumpOptimizer<'a> {
+    simple_func: &'a SimpleCfgFunction,
+}
+
+impl<'a> JumpOptimizer<'a> {
+    fn optimized(&mut self) -> SimpleCfgFunction {
```
ezrosent, line 22: Why have a separate struct / method here? Since there is only a single method for JumpOptimizer, would it be easier for this to be a standard function instead?

src/rvsdg/optimize_direct_jumps.rs
```diff
@@ -0,0 +1,113 @@
...
+        let mut resulting_graph: StableGraph<BasicBlock, Branch> = StableDiGraph::new();
+
+        // a map from nodes in the new graph to nodes in the
+        // old graph
+        // if a node was fused into another node, fix this
+        // so it points to the fused node
+        let mut node_mapping: HashMap<NodeIndex, NodeIndex> = HashMap::new();
+
+        let mut bfs = Bfs::new(&self.simple_func.graph, self.simple_func.entry);
```
ezrosent, line 31: nit: My read here is that we need BFS here to ensure that node_mapping[&previous] always succeeds. Maybe mention that here?

src/rvsdg/optimize_direct_jumps.rs
ezrosent: One possibility I can imagine is that "removing redundant copies" could get added to this same struct / maybe they could share some additional state. But that could just as easily be its own file.

ezrosent (changes requested):
Unit tests? I think it'd be helpful to run through cases like:
* graph where entry = exit
* graph with a node with a single outgoing edge to itself.

src/rvsdg/optimize_direct_jumps.rs
oflatt: So I've found myself spending a weird amount of time refactoring code to follow this pattern, and I think it's easier to set it up like this in the first place. For example, I did this to my RVSDG->CFG code, and a bunch of yihong's code.

src/rvsdg/optimize_direct_jumps.rs
oflatt: Yeah, good idea

oflatt:
Added unit tests for the cases you mentioned, good idea
Some awkwardness around importing the macros, see my TODO

src/rvsdg/optimize_direct_jumps.rs
```diff
@@ -0,0 +1,188 @@
...
+        return_ty: None,
+    };
+
+    cfg_test_equiv!(input_cfg.optimize_jumps(), []);
+}
+
+#[test]
+fn loops_to_self() {
+    // TODO you have to import both of these
+    //  is there a way to package up some macros all together?
```
ezrosent, line 150: You should be able to  put these macros in their own file and then use them in all these different test files. Is there a reason that wasn't working? (either way, I think you can do an absolute path rather than `super::super`.

ezrosent (approved):
Thanks! Happy to look at some of the macro scoping stuff in a follow up if it's not clear what to do.

src/rvsdg/optimize_direct_jumps.rs
oflatt: fixed!

PR #79, opened 2023-10-16
Expand constant folding to eliminate conditions

oflatt:
In this PR, I expand the constant folder to also eliminate conditions.
This involved several changes:

- adding proper booleans to egglog, and using them
- changing the encoding of bril ops like `add` to `badd` to avoid naming collisions with egglog
- adding an `OperandGroup` so that a `Gamma` can be unioned with a vector of outputs


src/rvsdg/from_egglog.rs
```diff
@@ -24,7 +24,25 @@ impl<'a> RvsdgFromEgglog<'a> {
...
-                    Operand::Project(*n as usize, self.egglog_term_to_body(body.clone()))
+                    let Term::App(bfunc, bargs) = body else {
+                        panic!("expected a body after project, got {}", self.termdag.to_string(body))
+                    };
+                    let bargs = bargs
+                        .iter()
+                        .map(|t| self.termdag.get(*t))
+                        .collect::<Vec<_>>();
+
+                    // Every `OperandGroup` should be used by a
```
oflatt, line 35: Could you actually drop a print edge here and fail to evaluate the print because we didn't evaluate the whole Operand Group?
Probably needs to be reflected in the rust RVSDG after all

oflatt: @yihozhang says this is fine because we always extract the state edge at the end.

src/rvsdg/egglog_optimizer/subst.egg
```diff
@@ -89,4 +90,4 @@
   )
   ( 
     (union (SubstVecVecOperandHelper (VVO vec) x v i) (VVO vec))
-  ) :ruleset subst)
+  ) :ruleset subst)
```
Alex-Fischman, line 93: What's the diff on this line? (GitHub issue not code review issue)

src/rvsdg/egglog_optimizer/mod.rs
```diff
@@ -1,21 +1,22 @@
...
+        subst_rules(),
         include_str!("shift.egg").to_string(),
         constant_fold_egglog(),
     ];
     code.join("\n")
 }
 
 pub fn rvsdg_egglog_schedule() -> String {
-    "(run 3)".to_string()
+    "(run-schedule (repeat 3 (run) (saturate subst)))".to_string()
```
Alex-Fischman, line 19: Can we document the schedule more? Specifically, I want to document all the problems that an arbitrary schedule would have, so that we can make sure that those problems don't come up as we change the schedule. For example, right now it should mention that subst is unextractable, so it needs to be saturated before extraction. (Also, do we need to saturate subst between each `run`?)

Alex-Fischman (changes requested):

src/rvsdg/egglog_optimizer/subst.egg
ajpal: it's just that there's no longer a newline at the end of the file
(we should probably add the newline back so that we don't have unnecessary diffs like this)

oflatt: fixed!

src/rvsdg/egglog_optimizer/mod.rs
oflatt: Good point- 
Right now, it's okay to not saturate subst- you can always extract the original program, of course.
This schedule just helps the subst rules saturate as an optimization. I'll add a comment.

Alex-Fischman (approved):
LGTM

PR #80, opened 2023-10-17
Add collatz.bril

clyben:
Add collatz.bril for running collatz conjecture. Also add Sub and Eq opeartion in Peg to pass test.


oflatt (changes requested):
Thanks for the benchmark!
You may have noticed giant snapshot tests- could you please merge main and re-generate those?

oflatt (approved):
Thanks!

PR #82, opened 2023-10-17
Don't interpret original program unless --interp passed

kirstenmg:
This PR changes `run` to only interpret the original program if the interp flag is passed. Before, it would interpret the original program but not the result program if the interp flag was not passed.


oflatt (approved):
Good change!

PR #83, opened 2023-10-17
Block params for gamma nodes should be live outputs, not live inputs.

### LINKED ISSUE #65: Block diamond is still not correct for rvsdg round trip
Printing out the value of "x" shows that we are still not doing RVSDG roundtrip correctly, even after #63

ezrosent:
Live outputs are the variables "used after" a block: that matches the variables that need to be passed to sub-blocks of a gamma node.

Fixes #65


oflatt (approved):
LGTM

PR #102, opened 2023-10-19
Add a repro test from #57

### LINKED ISSUE #57: RVSDG optimizer fails to share sub-loops

oflatt:
This closes #57 which was already fixed by #70 
It adds the benchmark from the issue.
I didn't want to add another snapshot test for this, but we will see in nightly if the cost is bad. I looked at the output and it's good.

### DIFF
--- a/tests/repro-loop-print.bril
+++ b/tests/repro-loop-print.bril
@@ -1,0 +1,11 @@
+@main() {
+  i: int = const 0;
+  one: int = const 1;
+  three: int = const 3;
+  .loop:
+    i: int = add i one;
+    cond: bool = lt i three;
+    br cond .loop .done;
+  .done:
+    print i;
+}


oflatt:
Oops, we already had this benchmark after all: `unroll_and_constant_fold.bril`. Closing

PR #106, opened 2023-10-19
Add loop unrolling to egglog

oflatt:
This PR implements loop unrolling for eggcc.
The rule is fairly simple, but requires a couple helpers. I've put those in `util.egg`.
Also, it requires 5 iterations to solve the benchmark we have, mostly because constant folding isn't in a separate ruleset yet.


src/rvsdg/egglog_optimizer/schema.egg
```diff
@@ -63,7 +63,7 @@
 (function Gamma (Operand VecOperand VecVecOperand) Body)
 ;; loops
 ;; predicate (inside loop), inputs (outside loop), outputs (inside loop)
-(function Theta (Operand VecOperand VecOperand) Body)
+(function Theta (Operand VecOperand VecOperand) Body :cost 20)
```
Alex-Fischman, line 66: Do other nodes have specific costs? Is the number 20 important? I can sort of guess what this does but a comment here would be good.

src/rvsdg/egglog_optimizer/subst.egg
```diff
@@ -91,3 +91,11 @@
   ( 
     (union (SubstVecVecOperandHelper (VVO vec) x v i) (VVO vec))
   ) :ruleset subst)
+ 
+
+;; ####################################
+;; implementation of PassThroughArguments
+;; Creates a vec of arguments
+;; (vec-of (Arg 0) (Arg 1) ...) with length i
+(function PassThroughArguments (i64) VecOperand :unextractable)
```
Alex-Fischman, line 100: Can you get rid of this stub by reordering the include_strs, or is there a circular dependency? (If there is a circular dependency please document it.)

src/rvsdg/egglog_optimizer/schema.egg
oflatt: Oops, I didn't mean to keep this change. I removed it, good catch.
In the future we will need a better cost function but not needed for this PR.

src/rvsdg/egglog_optimizer/subst.egg
oflatt: Moved to `util.egg` good catch

Alex-Fischman (commented):
LGTM

Alex-Fischman (approved):

PR #110, opened 2023-10-20
Disable Rvsdg SVG Snapshots

oflatt:
Rvsdg SVGs tend to be rather large, and nobody is looking at them.

I've disabled running them by default and simply added a few by hand.


Alex-Fischman (approved):
LGTM

oflatt:
I'm confused about the errors here and can't replicate them locally. Putting on hold for now.

PR #117, opened 2023-10-25
Fix function calls with return values

### LINKED ISSUE #116: Return values for function cause RVSDG -> CFG to fail
Failing example of a call of a function with a return value:
```
@main {
    a: int = const 0;
    c: int = call @inc a;
    print c;
}

@inc (x: int): int {
    one: int = const 1;
    res: int = add one x;
    ret res;
}
```

Fails with the following error:
```
thread 'main' panicked at 'Tried to unwrap state edge', src/rvsdg/to_cfg.rs:62:38
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Succeeding example:
```
@main {
    a: int = const 0;
    call @inc a;
}

@inc (x: int) {
    one: int = const 1;
    res: int = add one x;
    print res;
}
```

yihozhang:
Fixes #116

### DIFF
--- a/src/peg/mod.rs
+++ b/src/peg/mod.rs
@@ -141,7 +141,10 @@ impl PegBuilder<'_> {
                             ),
                             BasicExpr::Const(o, t, l) => BasicExpr::Const(*o, t.clone(), l.clone()),
                         };
-                        assert_eq!(0, selected);
+                        match expr {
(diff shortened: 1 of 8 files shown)


oflatt (approved):
Looks like a good fix to me

PR #119, opened 2023-10-25
fix conversion to egglog for (Node (OperandGroup ...))

yihozhang:
Once this gets in we can get function inlining working for some simple examples!

cc @kirstenmg


oflatt (approved):
Good fix

PR #120, opened 2023-10-25
Re-associate additions to promote constant folding

### LINKED ISSUE #93: Reassociate to promote constant folding
Example: `(+ 3 (+ x 4))` => `(+ (+ 3 4) x)`

Don't include Associative and commutative rules, instead canonicalize to one form

oflatt:
This PR re-associates additions, moving constants to the left.
Future work will include adding support for multiplication and pruning e-classes that are constant.

I think this is good enough to close #93 for our milestone.

### DIFF
--- a/src/rvsdg/egglog_optimizer/reassoc.rs
+++ b/src/rvsdg/egglog_optimizer/reassoc.rs
@@ -6,6 +6,7 @@
 pub(crate) fn reassoc_rules() -> String {
     let mut res = vec![];
     for op in ["badd", "bmul"] {
+        // Move constants to the left using commutativity
         res.push(format!(
             "
      (rule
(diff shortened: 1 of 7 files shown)


src/rvsdg/egglog_optimizer/reassoc.rs
```diff
@@ -0,0 +1,74 @@
...
+        .to_string(),
+    );
+
+    // Move constants up the tree
+    res.push(
+        "
+        (rule
+          ((= lhs (badd (IntT)
+                        a
+                        (Node (PureOp (badd (IntT) b c)))))
```
yihozhang, line 43: Just a note that this rule can blow up: consider `(+ n1 (+ n2 (+ n3 (+ n4 ...))))` for numbers `n1`, ..., where there are exponential ways to reassociate. And if any number is zero, this rule cannot even saturate.

yihozhang (approved):
Great work!

src/rvsdg/egglog_optimizer/reassoc.rs
oflatt: Yep, I had a comment on that

PR #121, opened 2023-10-25
Delete PEGs from codebase

### LINKED ISSUE #118: PEG cannot distinguish between different projections into a function call that has multiple return values
In #117, we have to disable [this check](https://github.com/oflatt/eggcc/blob/5757700b36c5798550ba338b9f3da9b554aba0ff/src/peg/simulate.rs#L155) for the following program with `--peg` flag.
```
@main {
    a: int = const 0;
    c: int = call @inc a;
    print c;
}

@inc (x: int): int {
    one: int = const 1;
    res: int = add one x;
    ret res;
}
```

This is because the state edge to `print` in this example is basically `(Project 1 (Call "inc" ...))` (obtaining the state edge out of the call), and in our current PEG implementation there is no way to distinguish this between `(Project 0 (Call "inc" ...))` which gets the actual return value.

Alex-Fischman:
This PR removes PEGs, because fixing them to handle multi-output calls correctly would be work that we don't need to do. Fixes #118. The tag peg-interpreter-still-worked saves the version of the code with the PEG stuff, in case we want it after our Bril deadline.


yihozhang (approved):
LGTM!

PR #122, opened 2023-10-27
Nightly Multiple Run Modes

ryan-berger:


infra/aggregate.py
```diff
@@ -0,0 +1,18 @@
+from sys import stdout
+from os import listdir
+from os.path import join
+import json
+
+root="./tmp/bench"
```
oflatt, line 6: ah, sweet sweet python
Please wrap the whole script in a function and call it at the bottom of the file for python reasons

infra/profile.sh
```diff
@@ -11,37 +11,48 @@ trap cleanup EXIT
...
-    # run eggcc, interp the program and put the data out to $out
-    out="./tmp/bench/${profile_name}.json"
+		# loop over RUNMODES and generate a profile for each, leaving it in the profile_name directory
+    for mode in ${RUNMODES[@]}
+		do
+		  echo "profiling $mode"
 
-    cargo run --release $profile --interp --profile-out "$out"
+			out="${mode}.json"
+    	pwd
```
oflatt, line 37: what is this pwd for? why is it so un-indented? XD

infra/profile.sh
```diff
@@ -11,37 +11,48 @@ trap cleanup EXIT
...
+			# generate the instruction count profile by interp'ing
+			cargo run --release $profile --interp --profile-out="./$out" --run-mode "$mode"
+			
+	    # $out now contains a key value of total_dyn_inst: value, so use read to get the key/value
+      # TODO: this is kind of a yaml sort of format so maybe yq would be good in the future
+			IFS=': ' read KEY VAL <<< $(cat $out)
```
oflatt, line 43: I think this line would be much clearer if you implemented in the python script

infra/profile.sh
```diff
@@ -11,37 +11,48 @@ trap cleanup EXIT
...
+		  echo "profiling $mode"
 
-    cargo run --release $profile --interp --profile-out "$out"
+			out="${mode}.json"
+    	pwd
+			# generate the instruction count profile by interp'ing
+			cargo run --release $profile --interp --profile-out="./$out" --run-mode "$mode"
+			
+	    # $out now contains a key value of total_dyn_inst: value, so use read to get the key/value
+      # TODO: this is kind of a yaml sort of format so maybe yq would be good in the future
```
oflatt, line 42: comments indented weirdly

infra/profile.sh
```diff
@@ -11,37 +11,48 @@ trap cleanup EXIT
...
+			# generate the profile and overrite the $out file
+			hyperfine --warmup 2 --export-json "$out" "cargo run --release $profile --interp --run-mode $mode"
 
-    # export hyperfine out to tmp file
-    hyperfine_out="./tmp/hyperfine/${profile_name}.json"
-    hyperfine --warmup 2 --export-json "$hyperfine_out" "cargo run --release $profile --interp"
+			# overwwrite outfile with json version of profile data, annotate with profile name.
+    	# we also combine both instruction count and hyperfine json output into a single object
+    	# to make things super easy
+			printf '{"%s": "%s", "hyperfine": %s}' \
```
oflatt, line 51: would prefer you generate a file with a different name instead of overwriting

infra/profile.sh
```diff
@@ -11,37 +11,48 @@ trap cleanup EXIT
...
+    profile="../../../$1"
 
     # strip the file path down to just the file name
     # TODO: profile name is not unique, generate a unique output path (it will be aggregated anyway)
-    profile_name=$(basename -- $profile)
+    profile_file=$(basename -- $profile)
+    profile_name="${profile_file%.*}"
+
+    mkdir ./tmp/bench/"$profile_name"
+    pushd ./tmp/bench/"$profile_name"
```
oflatt, line 29: Instead of using pushd, can the whole bash script stay in the same directory but use this path for `$out`?

oflatt (changes requested):
Awesome work, lots of minor things with the bash.
I'm not a bash expert but I put comments on things I would change

infra/profile.sh
ryan-berger: hahah, `vim` totally messed everything up... I didn't notice until now, let me run a formatter

infra/aggregate.py
ryan-berger: Sure. Do you want me to do the whole `if __name__ == 'main'`?

infra/profile.sh
ryan-berger: :( ok

infra/profile.sh
ryan-berger: If I'm gonna do that `read` in Python, then yes, I'll want a different file

infra/profile.sh
ryan-berger: debugging, and it is indented because I don't know how to use vim

infra/aggregate.py
oflatt: Nah just a function is good

ryan-berger:
@oflatt I've made all the changes you've requested and also added in a bunch more benchmark files :). A re-review would be lovely

infra/aggregate.py
```diff
@@ -0,0 +1,49 @@
+from sys import stdout
+from os import listdir
+from os.path import join
+import json
+
+
+# main will aggregate all profile info into a single json array.
```
oflatt, line 7: Great comment

infra/profile.sh
```diff
@@ -9,45 +9,41 @@ cleanup() {
 trap cleanup EXIT
 
 # TODO: take in file glob as command line argument
-PROFILES=tests/small/add.bril
+PROFILES=(tests/small/*.bril)
```
oflatt, line 12: This is fine for now, but we actually test on files outside the `small` folder as well. Can be fixed in a future PR or on this one.
The current `files.rs` testing harness actually just looks recursively for bril files, excluding ones in with `failing` or `should_fail` in the full path. Which is a hack

oflatt (approved):
Lovely, this is a big improvement. See comment on benchmark list

infra/profile.sh
ryan-berger: Yeah, I'm not particularly happy with how we currently do it, we shouldn't need a high level language to decide whether we should exclude tests, it should be super easy to glob all our benchmarks.

I think we probably want to have actual benchmarks rather than tests or else we'll get a lot of pollution on our tables, which should help. We could put those in a bench/ folder or something

PR #123, opened 2023-10-27
Function inlining

kirstenmg:
Add the function inlining optimization.


kirstenmg:
Currently, for the recursive example, the call doesn't get inlined at all. This is not expected, I would expect it to unroll a few calls based on how many times we apply our rules.

kirstenmg:
The reason that recursive function calls do not get inlined is that the function body also has a function call, so inlining only increases the cost of the rewritten program (thank you Oliver for figuring this out)

oflatt (approved):
This is great stuff- love how simple the inlining rule is.

PR #127, opened 2023-10-29
Move simultaneous substitutions into pure egglog

ezrosent:
This also fixes an issue where the initial "nested" implementation was not capture-avoiding.


oflatt (approved):
Great work! This fixes the bug and I like the code more.

PR #136, opened 2023-10-31
Exclude loop back-edges when finding join points.

### LINKED ISSUE #130: Flatten loop benchmark crashes in cfg->rvsdg translation
https://github.com/oflatt/eggcc/blob/oflatt-flatten-loops/tests/small/flatten_loop.bril

ezrosent:
We were getting inconsistent answers for join points because a loop back-edge was "tricking" the join point algorithm into stopping early. The fix is to just filter out back-edges using the usual algorithm of scanning the dominators.

Fixes #130


oflatt (approved):
Nice fix

PR #137, opened 2023-10-31
for one visualization, don't print annoying text

oflatt:


ajpal (approved):

src/main.rs
```diff
@@ -57,6 +57,8 @@ fn main() {
 
     if args.interp {
         println!("{}", result.result_interpreted.unwrap());
+    } else if let &[visualization] = &result.visualizations.as_slice() {
+        println!("{}", visualization.result);
```
ajpal, line 61: Maybe a comment indicating we're omitting the text below just as a convenience?

PR #138, opened 2023-10-31
Interval Analysis for Gamma Elimination

ajpal:


oflatt (approved):
Looks like a great start to interval analysis, and it's working!
We will have to be careful when we start to write rules over integers and stuff

Issue #135, opened 2023-10-30
Zach's cool if rewrite


oflatt:
negate ifs- flip condition and swap then and else

de morgan's laws

Alex-Fischman:
Should this be closed? It's in the done column on the board...

oflatt:
I was hoping @ajpal would close and move it, trying to get people to use the board

PR #139, opened 2023-10-31
Add another gamma rewrite rule

ajpal:
`if a then A else B ==> if !a then B else A`


oflatt (approved):
Nice rule!
I think there are several others in issue #135 
Looks like you need to do `cargo insta review`

oflatt (approved):
these rules rule

PR #141, opened 2023-10-31
Range check optimization

### LINKED ISSUE #97: Remove range checks in loops
From LLVM:
https://llvm.org/doxygen/InductiveRangeCheckElimination_8cpp_source.html

Alex-Fischman:
This PR implements a range check optimization, shown in pseudocode below
```
i = 0
do:
 if i < 5:
   print 1
 else:
   print 2
while (i < 10)
print i
--------------
i = 0
do:
 print 1
while (i < 5)
if (i < 10):
 do:
   print 2
 while (i < 10)
print i
```
Right now this should not be merged because I don't know how to test it (unlike the gamma rewrites, this rewrite is probably a speed improvement, but it's bad for code size so we never extract it).
Fixes #97.


Alex-Fischman:
Closing this and opening a new PR on a different branch.

PR #142, opened 2023-11-02
Add purity analysis

rtjoa:
Add `(relation TYPE-is-pure (TYPE))` where `TYPE` is one of `{Expr, Body, Operand, VecOperand, VecVecOperand}`.

We consider something pure if interpreting it wouldn't print anything, e.g. a function that passes around a print edge but never uses it would still be considered pure.

We also do some minor refactoring, moving vector reification into the `fast-analyses` ruleset - meant for frequently-used, at-most-linear-time analyses (e.g. this one).


oflatt (approved):
Great work! Looks like you have some good tests as well.

PR #143, opened 2023-11-02
Add `Body-contains-*` rules

rtjoa:
Add `(relation Body-contains-TYPE (Body i64 TYPE))` where `TYPE` is one of `{Expr, Body, Operand}`. If we call the arguments `body`, `i`, and `e`, a tuple of one of these relations indicates that `body` contains `e` in its `i`th output.

A Body is considered to contain something if it's in the Body's context. Notably, if a Gamma/Theta contains `x`, then the Args in `x` must be bound by that Gamma/Theta.

To get the intuition:
- A region containing a Gamma will contain that Gamma's inputs and predicate.
- A region containing a Theta will contain that Theta's inputs.
- A Gamma contains its own outputs.
- A Theta contains its own predicate and outputs.

### DIFF
--- a/src/rvsdg/egglog_optimizer/fast_analyses.rs
+++ b/src/rvsdg/egglog_optimizer/fast_analyses.rs
@@ -155,14 +155,15 @@ fn is_pure_rules() -> Vec<String> {
     res
 }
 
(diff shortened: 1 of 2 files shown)


oflatt (approved):
Nice- good comments

PR #144, opened 2023-11-03
Conditional invariant code motion

rtjoa:
## Summary

If all branches of a `Gamma` share code, we can compute it outside the body and pass the result via a new argument.

Example:
| Unoptimized  | Optimized |
| ------------- | ------------- |
| image | image |

As this uses `can-subst-beneath`, which is slow/doesn't saturate, we move it to its own ruleset, `subst-beneath`, and modify the schedule to use it more sparingly. In the future we hope to improve `can-subst-beneath` (contextual equality, named wires), separately from this PR.

## Testing
Added Bril test at `tests/small/branch_duplicate_work.bril` and RVSDG-level tests at `src/rvsdg/tests.rs`.

### DIFF
--- a/src/rvsdg/egglog_optimizer/conditional_invariant_code_motion.egg
+++ b/src/rvsdg/egglog_optimizer/conditional_invariant_code_motion.egg
@@ -1,0 +1,45 @@
+; If (= x (prefix-containing-operand (Gamma ...) e)), then the first x outputs
+; of Gamma contain e.
+(function prefix-containing-Operand (Body Operand) i64 :merge (max old new))
+(rule
+    ((= gamma (Gamma pred inputs outputs))
+     (Body-contains-Operand gamma 0 e))
+    ((set (prefix-containing-Operand gamma e) 0))
+    :ruleset fast-analyses)
+(rule
+    ((= gamma (Gamma pred inputs outputs))
+     (= i (prefix-containing-Operand gamma e))
+     (Body-contains-Operand gamma i e))
+    ((set (prefix-containing-Operand gamma e) (+ i 1)))
+    :ruleset fast-analyses)
+
+; Limit blowup by only repeatedly hoisting so many times
+(let max-hoists 1)
+(function hoist-count (Body) i64 :merge (max old new))
+(rule ((= gamma (Gamma pred inputs outputs)))
+      ((set (hoist-count gamma) 0))
+      :ruleset fast-analyses)
+
+; Hoist code out of a gamma, adding it as an argument instead.
+(rule 
+    ((= gamma (Gamma pred (VO inputs) outputs))
+     ; All branches contain hoist-me
+     (= (prefix-containing-Operand gamma hoist-me)
+        (VecVecOperand-length outputs))
+     (!= hoist-me (Arg any))
+     ; Can't add another version of an impure node since we might extract both.
+     ; Note: if we lift this requirement via an extraction-like substitution
+     ; procedure, then remember to add (Operand-is-pure hoist-me)!
+     (Body-is-pure gamma)
+     (= n (VecOperand-length (VO inputs)))
+     (< (hoist-count gamma) max-hoists)
+     (> (VecVecOperand-length outputs) 1))
+    ((let new-inputs
+        ; Move hoist-me into the parent context and add it as an input
+        (VO (vec-push inputs (SubstOperandAll hoist-me (VO inputs)))))
+     (let new-gamma (Gamma pred new-inputs outputs))
+     (union gamma new-gamma)
+     ; Allow the hoisted code to be replaced with the new argument
+     (can-subst-Operand-beneath (GammaCtx new-inputs) hoist-me (Arg n))
+     ; Increment hoist count
+     (set (hoist-count new-gamma) (+ 1 (hoist-count gamma)))))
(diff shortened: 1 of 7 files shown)


src/rvsdg/egglog_optimizer/conditional_invariant_code_motion.egg
```diff
@@ -0,0 +1,50 @@
...
+    :ruleset fast-analyses)
+
+; Limit blowup by only repeatedly hoisting so many times and by limiting the
+; number of branches we hoist from or max number of inputs we can add.
+(let max-hoists 1)
+(function hoist-count (Body) i64 :merge (max old new))
+(rule ((= gamma (Gamma pred inputs outputs)))
+      ((set (hoist-count gamma) 0))
+      :ruleset fast-analyses)
+(let max-branches 2)
```
oflatt, line 23: What's the point of max-branches and max-inputs? Doesn't max-hoists take care of the blowup?

src/rvsdg/egglog_optimizer/mod.rs
```diff
@@ -37,11 +38,16 @@ pub fn rvsdg_egglog_schedule() -> String {
...
 
-    "(run-schedule
+    "
+    (run-schedule
         (repeat 5 (saturate fast-analyses)
                   (run)
-                  (saturate subst)))"
-        .to_string()
+                  (saturate subst))
+        (repeat 4 subst-beneath)
```
oflatt, line 46: Ooh can't say I love this schedule- does it have to be this complex, with constants like 4 and 2?
Is this because we are not allowed to saturate subst-beneath because it might not terminate?

oflatt: Also is this indentation weird?

oflatt (changes requested):
Had a couple questions

src/rvsdg/egglog_optimizer/conditional_invariant_code_motion.egg
rtjoa: To limit the number of Gammas we try to hoist in the first place, sadly this these are also necessary

rtjoa: I think the tradeoff is --
lower limits -> less performance interference with other's optimizations
higher limits -> people can more easily write rules relying on CICM

Because our performance is very bad now, I chose the former

src/rvsdg/egglog_optimizer/mod.rs
rtjoa: Changed it to be less complex:
```
        (repeat 5 (saturate fast-analyses)
                  (run)
                  (saturate subst))
        (repeat 6 subst-beneath (saturate fast-analyses))
```

The indentation is meant to have future changes produce less diff. Just made it a bit less weird.

src/rvsdg/egglog_optimizer/conditional_invariant_code_motion.egg
oflatt: Hmm that's unfortunate. Okay
@clyben is experimenting with an analysis that finds nodes on the boundary between invariant and not invariant, and uses those. But okay for this PR

oflatt (approved):
Great PR, really pushing what we can do right now. I'd say this is complete for our deadline next week, lets return to this question after.

src/rvsdg/egglog_optimizer/conditional_invariant_code_motion.egg
```diff
@@ -0,0 +1,50 @@
...
+(let max-inputs 4)
+
+; Hoist code out of a gamma, adding it as an argument instead.
+(rule 
+    ((= gamma (Gamma pred (VO inputs) outputs))
+     ; All branches contain hoist-me
+     (= (prefix-containing-Operand gamma hoist-me)
+        (VecVecOperand-length outputs))
+     (!= hoist-me (Arg any))
```
yihozhang, line 32: The semantics here is not what you intended. Here, if there exists an `any` such that `(!= host-me (Arg any))`, this check will pass. @clyben is working on something like a `not-an-arg` analysis that will be helpful

src/rvsdg/egglog_optimizer/conditional_invariant_code_motion.egg
```diff
@@ -0,0 +1,50 @@
...
+     ; Can't add another version of an impure node since we might extract both.
+     ; Note: if we lift this requirement via an extraction-like substitution
+     ; procedure, then remember to add (Operand-is-pure hoist-me)!
+     (Body-is-pure gamma)
```
yihozhang, line 36: Why does `gamma` (and `hoist-me`) has to be true? Even if `hoist-me` is not true, couldn't you still push them out?

rtjoa: Do you mean pure?

I think hoist-me has to be pure, because can-subst-beneath is allowed to replace some but not all, but I think we can drop the requirement that gamma is pure.

yihozhang: Makes sense. Thanks!

src/rvsdg/egglog_optimizer/conditional_invariant_code_motion.egg
oflatt: did this get resolved?

rtjoa: No, although this isn't a soundnes problem, it will just cause some useless single-arg hoists. Once @clyben merges `not-an-arg` I'll be sure to update.

PR #145, opened 2023-11-03
Sweet Sweet Pretty Printing

oflatt:


oflatt:
@yihozhang I had to remove your bad test.
You were using `process_commands` instead of `run_commands` so the test wasn't testing anything!

oflatt:
Added issue #148

yihozhang (approved):
Looks good!

PR #147, opened 2023-11-03
Push intervals in and out of Gamma contexts

ajpal:
Note Oct 31, 2023


src/rvsdg/egglog_optimizer/interval-analysis.egg
```diff
@@ -65,3 +76,74 @@
...
+(rule (
+      (= gamma (Gamma pred (VO ins) (VVO outs)))
+      (NumN n)
+      (= arg (vec-get ins n))
+      (= (IntI l h) (ival arg))
+      )
+      ((set (context-ival (Arg n) gamma) (IntI l h)))
+)
+
+; interval of a + b = (la + lb, ha + hb)
```
ajpal, line 98: @oflatt it's pretty painful to have 4 versions of this rule- is there a way to match on "Either an ival or a context-ival"?

src/rvsdg/egglog_optimizer/mod.rs
```diff
@@ -31,7 +31,7 @@ pub fn rvsdg_egglog_schedule() -> String {
     // they take many iterations.
 
     "(run-schedule
-        (repeat 5 (saturate fast-analyses)
+        (repeat 7 (saturate fast-analyses)
```
ajpal, line 34: @oflatt is it ok to bump this?

src/rvsdg/egglog_optimizer/interval-analysis.egg
```diff
@@ -65,3 +76,74 @@
         (union gamma (OperandGroup (SubstVecOperandAll (vec-get outputs 0) inputs)))
       )
 )
+
+; HACK: Table of fucking integers
```
oflatt, line 80: I would also put a TODO 
 @rtjoa I think you are working on table for ranges of integers?

src/rvsdg/egglog_optimizer/interval-analysis.egg
oflatt: One way to do it would be to have another table called `either-ival-or-context-ival` and a rule that adds to that table. I think that would be reasonable 
Another strategy is to generate these rules from rust.

src/rvsdg/egglog_optimizer/mod.rs
oflatt: Instead, could you please add your interval rules to the `fast-analyses` ruleset?

oflatt (changes requested):
Great work! Minor changes

src/rvsdg/egglog_optimizer/interval-analysis.egg
ajpal: What would the type of this table be? Are there `Option`s?

src/rvsdg/egglog_optimizer/interval-analysis.egg
rtjoa: I was not, but I just talked to Anjali and it seems like `VecOperand-get` is sufficient for this PR.

The range rules would be straightforward to write, but I'm unsure if we need them in general.

oflatt (approved):
Great! remove output.egg and we are good to go

output.egg
```diff
@@ -0,0 +1,1325 @@
+(datatype Literal)
```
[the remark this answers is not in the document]
ajpal: oops, sorry!

PR #149, opened 2023-11-06
Fix rvsdg_odd_branch_egg_roundtrip test

### LINKED ISSUE #148: Fix deleted test in pretty printing PR
Pr #145 deletes a test which was doing nothing
We should fix this test

yihozhang:
Fixes #148 so that it is not deleted.


oflatt (approved):
Great!

PR #153, opened 2023-11-07
Loop strength reduction

kirstenmg:
Working specific example of loop strength reduction. Needs to be made general.


kirstenmg:
Will likely not work for all cases, due to cost model issues (optimized loop body is over-counted since it uses an OperandGroup of Projects).

src/rvsdg/egglog_optimizer/mod.rs
```diff
@@ -35,17 +36,18 @@ pub fn rvsdg_egglog_schedule() -> String {
     "(run-schedule
         ; It is sound to not saturate fast-analyses/subst, but we do because
         ; they won't blow up and will help other rules go through.
-        (repeat 5
-            (saturate fast-analyses)
-            (run)
-            (saturate subst))
+        (repeat 2
+            (run loop-strength-reduction)
```
oflatt, line 40: I don't think we need a loop-strength-reduction ruleset. Just add to the normal ruleset (don't include a ruleset on your rules)

src/rvsdg/egglog_optimizer/loop_strength_red.egg
```diff
@@ -0,0 +1,76 @@
...
+;; c = 3
+;; d = 0
+;; for i = 0 to n:
+;;     a += d
+;;     d += c
+
+(ruleset loop-strength-reduction)
+
+;; For a      theta, operand  is in the theta and either invariant or constant
+(relation inv (Body Operand))
```
oflatt, line 20: We already have an invariant analysis in `loop_invariant.egg` that you can reuse I think.

src/rvsdg/egglog_optimizer/loop_strength_red.egg
```diff
@@ -0,0 +1,76 @@
...
+        (Body-contains-Operand old-theta contains-i old-exp)
+    )
+    (
+        ; Each time we need to update d by the product of the multiplied constant and the loop increment
+        (let addend (Node (PureOp (bmul (IntT) c loop-incr))))
+        ; n is index of our new, temporary variable d
+        (let n (vec-length inputs))
+
+        ; Initial value of d is i * c
+        (let d-init (Node (PureOp (bmul (IntT) c (VecOperand-get (VO inputs) i)))))
```
oflatt, line 55: Did you test this on an invariant argument instead of a constant? Doesn't it seem like you need to substitute here since the initial value is outside the loop?

oflatt (changes requested):
Great work! Had some small changes

src/rvsdg/egglog_optimizer/loop_strength_red.egg
kirstenmg: Yes, invariant argument does not work currently.

src/rvsdg/egglog_optimizer/mod.rs
kirstenmg: Removed

PR #154, opened 2023-11-07
Remove references to arguments that are passed through if statements and loops

### LINKED ISSUE #113: Remove unnecessary ports

oflatt:
Fixes #113


ezrosent:
Very interesting. A couple high-level questions to help my understanding (can discuss offline too):

* Do we want to make this our eventual method for extracting everything as well?
* If not, do these optimizations still work? In the first impure rule:
```clojure
;; If a theta passes along argument,
;; can extract the input instead.
(rule ((= lhs (Project index loop))
       (= loop (Theta pred inputs outputs))
       (= (VecOperand-get outputs index) (Arg index))
       (= passedthrough (ExtractedOperand (VecOperand-get inputs index))))
      ((set (ExtractedOperand lhs) passedthrough)))
```
Seems to just modify the `ExtractedOperand` rule, which gives you a `TermAndCost`, and then we have a separate rule that "reflects" extracted terms back into normal terms:

```clojure
(rule ((= theta (Theta pred inputs outputs))
       (= (BodyAndCost extracted cost)
          (ExtractedBody theta)))
      ((union theta extracted)))
```

But I'm not sure how this helps with linearity: It seems like in order to ensure an effect is performed the same number of times pre-and post-rewrite, you would also have to modify `loop` to have fewer outputs, and route loop-varying projections to the new loop. But I don't see any loop splitting anywhere (does that happen elsewhere?).

* It seems like the "more idiomatic" way to handle linearity issues with impure projections would be to "glue" the rerouted edge to the old loop: here's some pseudocode
```clojure
; loop has one output that is invariant and unused within `pred` and the other outputs.
(let loop (Theta pred ins outs))
(let output (vec-get ins 1)) ; the invariant output
; smaller-loop is the same loop with everything remapped, but with `output` removed
(let smaller-loop (Theta pred' ins' outs'))
(union loop (OperandGroup (vec-of (Proj 0 smaller-loop) output (Proj 1 smaller-loop) ...)))
```

This way, `(Proj 1 loop)` gets routed to `output` so long as the smaller loop has lower cost. Maybe that's implicitly happening here, but "unused" determinations are hard to do without fixing a term via extraction?

src/rvsdg/egglog_optimizer/extraction_rules.rs
```diff
@@ -0,0 +1,204 @@
+use super::BRIL_OPS;
+
+/// These extraction rules are not (yet) used to perform actual extraction.
```
ezrosent, line 3: I think we need more text here. 

In particular, we do the optimization _during_ extraction, we don't have subset of rules that simply look at the extractions output.

src/rvsdg/egglog_optimizer/passthrough_optimize.rs
```diff
@@ -30,13 +54,27 @@ pub(crate) fn passthrough_optimize_rules() -> String {
...
-       ;; also subsume the project
-       (delete (Project index loop)))
-      :ruleset {ruleset})
-        "
-    ));
+       (= passedthrough (ExtractedOperand (VecOperand-get inputs index))))
+      ((set (ExtractedOperand lhs) passedthrough)))
+
+
+;; if we reach a new context, union
```
ezrosent, line 61: I think we need more documentation of these three rules as well.

ezrosent (changes requested):
We discussed a bunch of the stuff in my previous comment offline. I think I'm convinced we should go through with this approach, but two questions remain


1. The standard line that I've heard about "linearity" issues around state edges is that we need to avoid extracted terms using state edges more times than they were in the original program. The way we do this is to ensure that a given state edge always "lives in the same eclass it started in." This PR breaks that slightly: if the invariant edge of a gamma or theta region is a state edge, that state edge will still be passed to the old region in which it is invariant, it just will not be used. I think we need to decide if this is okay (if it isn't, we need to build a version of the inner region without that edge, and that's a bit harder).
2. Re: the approach sketched in my first comment, Oliver pointed out that you want to avoid rules having to pattern match on a bunch of nested `Project`s. I think you can flatten n of these `Project`s into 1, but going from 1 to 0 probably runs into linearity problems: but aren't all of our pattern matches in terms of projects already?

oflatt:
@ezrosent I improved performance of this PR and improved the comments.
To answer your questions.

1. I think it would be much cleaner to remove the old edges, but probably not worth the effort for eggcc right now. Here's an argument for why it's still sound for now:

The purpose of print edges is to enforce an ordering on when the prints occur. To do this, we pass the print state through linearly, with one producer and one consumer. This optimization breaks that- there are multiple consumers of the print state after the optimization. Luckily, the output port with the print edge from the context will never be used-it will remain dead. So my argument is that having multiple references to the print edges is okay, as long as only one is allowed to actually print and pass it along.

2. My problem with your first comment isn't the addition of more `Project` statement, my problem is that we now have an `OperandGroup` in the way. Multiple `OperandGroup` cannot be collapsed easily- it would require something like what I have in this PR.

ezrosent (approved):

PR #155, opened 2023-11-07
Have `can-subst-beneath` take `Context` rather than `Body`

rtjoa:
Rather than an abritrary body, have `above` in `can-TYPE-subst-beneath` (#111) take a `Context`:
```
(datatype Context
   (GammaCtx VecOperand)
    (ThetaCtx VecOperand))
```

Where one would previously write:
`(can-subst-TYPE-beneath (Gamma pred inputs outputs) from to)`
One now writes:
`(can-subst-TYPE-beneath (GammaCtx inputs) from to)`

And likewise for `Theta`. This allows some to subst beneath one body but not another if they differ in inputs, even if they are unioned.


oflatt (approved):
Looks good to me- we will have to use this carefully, and might want more types of context in the future

PR #158, opened 2023-11-08
Optimize passed through arguments, but only in the pure case

oflatt:
This PR is blocked on subst rules not saturating when there is a cycle.

It also relies on the argument that unioning a project with something else is valid when the whole thing is pure.
The non-pure version is much harder.

### DIFF
--- a/src/rvsdg/egglog_optimizer/loop-optimizations.egg
+++ b/src/rvsdg/egglog_optimizer/loop-optimizations.egg
@@ -8,7 +8,7 @@
        (let pass-through (PassThroughArguments (vec-length outputs)))
        (union theta
           (Gamma
-            (SubstOperandAll pred after-one-iter)
+            (SubstOperandAll pred (VO inputs))
             after-one-iter
             (VVO
               (vec-of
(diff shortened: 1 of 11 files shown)


src/rvsdg/egglog_optimizer/loop-optimizations.egg
```diff
@@ -8,7 +8,7 @@
        (let pass-through (PassThroughArguments (vec-length outputs)))
        (union theta
           (Gamma
-            (SubstOperandAll pred after-one-iter)
+            (SubstOperandAll pred (VO inputs))
```
ezrosent, line 11: I agree this one is correct; just confirming the fix is intended?

src/rvsdg/egglog_optimizer/passthrough_optimize.rs
```diff
@@ -0,0 +1,42 @@
...
+       (= (VecOperand-get outputs index) (Arg index))
+       (= passed-through (VecOperand-get inputs index))
+       (Body-is-pure loop)
+      )
+      ((union lhs passed-through)
+       ;; also subsume the project
+       (delete (Project index loop)))
+      :ruleset {ruleset})
+
+;; If a gamma passes along an argument in both branches,
```
ezrosent, line 22: nit, "if a binary gamma"

ezrosent (approved):

src/rvsdg/egglog_optimizer/loop-optimizations.egg
oflatt: Yep, forgot to mention I fixed this!

PR #161, opened 2023-11-08
Add bool-= egglog opeartion

clyben:
Minor changes with adding bool-=
Block by issue #164


oflatt:
Is this still relevant? We should merge or close

clyben:
I think no if we are not using this encoding.

PR #163, opened 2023-11-08
All Nightlies

### LINKED ISSUE #132: Run all benchmarks in nightly

ryan-berger:
Fixes #132

### DIFF
--- a/infra/profile.sh
+++ b/infra/profile.sh
@@ -9,7 +9,7 @@ cleanup() {
 trap cleanup EXIT
 
 # TODO: take in file glob as command line argument
-PROFILES=(tests/small/*.bril)
+PROFILES=(tests/*.bril tests/small/*.bril tests/brils/passing/**/*.bril)
 
 RUNMODES=("nothing" "rvsdg-optimize")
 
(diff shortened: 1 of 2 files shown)


infra/profile.sh
```diff
@@ -9,7 +9,7 @@ cleanup() {
 trap cleanup EXIT
 
 # TODO: take in file glob as command line argument
-PROFILES=(tests/small/*.bril)
+PROFILES=(tests/*.bril tests/small/*.bril tests/brils/passing/**/*.bril)
```
oflatt, line 12: Can you just loop recursively though the directories, but don't do paths that include `failing` or `should_fail` or `snapshots`?
Or is that beyond bash abilities lol
Should have written in python (tm)

oflatt (changes requested):

infra/profile.sh
ryan-berger: I'd like to see this as a temporary solution to the problem of getting good benchmarks rather than investing the effort to skip should_fail/fail/snapshots and having to tear it out later.

I'm sure that bash could do this, but I don't think that 3 globs are too hard to maintain over having to maintain a blacklist of 3 folders.

PR #166, opened 2023-11-09
Loop invariant motion

clyben:
This PR add loop invariant motion.
In a Theta node, an operand is movable out of Theta when 
1. It's an invariant operand, and is on boundary of invariant and non-invariant. (e.g. op1 is invariant but (+ op1 op2) is not invariant, then op1 is on boundary)
2. The operand is not basic input like (Arg n), or otherwise it's not needed to move.
Currently it's facing performance issues which are likely caused by substitution. It might be solved with must-subst-beneath.


clyben:
currently it's wouldn't generate the wanted code due to the OperandGroup extraction problem.

src/rvsdg/egglog_optimizer/loop-optimizations.egg
```diff
@@ -1,12 +1,15 @@
-
+(function unroll_count (i64) i64 :unextractable :merge (max old new))
```
oflatt, line 1: Do you still need this?

oflatt (changes requested):
Had a quick question

src/rvsdg/egglog_optimizer/loop-optimizations.egg
clyben: yes, setting the unroll max to 2 will make the test not finishing..

oflatt: Okay, fix after today!

oflatt (approved):
Great work!

PR #167, opened 2023-11-10
Range splitting

Alex-Fischman:
This PR implements a range check optimization, shown in pseudocode below
```
i = 0
do:
 if i < 5:
   print 1
 else:
   print 2
while (i < 10)
print i
--------------
i = 0
do:
 print 1
while (i < 5)
if (i < 10):
 do:
   print 2
 while (i < 10)
print i
```
Right now this should not be merged because I don't know how to test it (unlike the gamma rewrites, this rewrite is probably a speed improvement, but it's bad for code size so we never extract it).
Fixes https://github.com/oflatt/eggcc/issues/97.


oflatt (approved):
Great! This closes #97 
Assigned you to generalize this, #170

PR #168, opened 2023-11-10
Fix error handling to print proper error message instead of debug struct

oflatt:


ezrosent (approved):
Optional: 

Given that we now have multiple pieces of code like:

```rust
match result { 
Ok(result) => result,
Err(error) => { panic!("{}", error) }
}
```

I think we should consider giving `EggCCError` a better `Debug` implementation. Then, we could just call `expect` or `unwrap` on these results, which would make the code a bit cleaner.

oflatt:
Will fix EggCCError sometime, merging so people get error message now

PR #169, opened 2023-11-11
Add print rules to can-subst-beneath

kirstenmg:
`can-subst-*-beneath` would get stopped by a PRINT because there was no rule to propagate it through a print.


oflatt (approved):
lgtm

PR #178, opened 2023-11-13
Optimize `can-subst-beneath` by checking whether `new-from` is in `above` before creating new tuple

rtjoa:
Before, 20 iterations of `subst-beneath` OOM'd after 15min on my computer. With this change, it takes ~30 seconds.

We conservatively increase the number of iters from 6 to 10.

Thanks to @yihozhang for originally suggesting this optimization!

### DIFF
--- a/src/rvsdg/egglog_optimizer/mod.rs
+++ b/src/rvsdg/egglog_optimizer/mod.rs
@@ -43,7 +43,7 @@ pub fn rvsdg_egglog_schedule() -> String {
         ; spine - we are working on this!), so we only run it a few times at the
         ; end to apply substitutions that the main optimizations find. It's
         ; interleaved with fast-analyses because it relies on reified vecs.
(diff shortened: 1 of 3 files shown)


oflatt (approved):
Looks good to me, nice optimization.

src/rvsdg/egglog_optimizer/mod.rs
```diff
@@ -43,7 +43,7 @@ pub fn rvsdg_egglog_schedule() -> String {
         ; spine - we are working on this!), so we only run it a few times at the
         ; end to apply substitutions that the main optimizations find. It's
         ; interleaved with fast-analyses because it relies on reified vecs.
-        (repeat 6 subst-beneath (saturate fast-analyses))
+        (repeat 10 subst-beneath (saturate fast-analyses))
```
oflatt, line 46: Is it fast enough we can run this during normal rule execution, not just at the end?

PR #179, opened 2023-11-14
Use a greedy dag extraction algorithm

oflatt:
I've implemented greedy dag extraction for egglog here:https://github.com/egraphs-good/egglog/pull/290
It's significantly slower than normal extraction, since it doesn't do any fancy tricks like use a persistent data structure.

This helps overcome problems with extraction the come up, for example in #153


oflatt:
People should feel free to try this out- not quite ready to merge

oflatt:
This PR now uses a new feature I added to my egglog branch: `:context-wrapper` annotations.
To do this, I did a refactor to add a new `VecOperandCtx` constructor for `Gamma` to use.

yihozhang (approved):
lgtm

PR #180, opened 2023-11-14
Gamma Pull In

ajpal:
```
if C:
  S1
else:
  S2
S3
-----
if C:
  S1; S3
else:
  S2; S3
```

Right now, special-cased to Plus
Unions a Plus of Projects with a Project of a new Gamma with an output port added that does the Plus within the body of each branch.

Note Nov 8, 2023


oflatt (approved):
Great work!

PR #182, opened 2023-11-15
Fix De Morgan's laws

Alex-Fischman:
Fix a bug in the rewrite rules and make them operate directly on boolean expressions instead of gamma nodes.


oflatt (approved):
great

PR #185, opened 2023-11-15
Nightly Comparisons

ryan-berger:


infra/nightly-resources/index.js
```diff
@@ -1,46 +1,202 @@
...
-
-  let container = document.getElementById("profile");
-  let parsed = [];
-
-  // aggregate benchmark runs into a list by their "benchmark" key
-  let groupedBy = {};
-  data.forEach((obj) => {
-    if(!groupedBy[obj.benchmark]) {
-      groupedBy[obj.benchmark] = [];
+async function getPreviousRuns() {
```
oflatt, line 1: nice- I would add a todo to fix this when we have a good way to get it cleanly

infra/nightly-resources/index.js
```diff
@@ -1,46 +1,202 @@
...
+    select.onchange();
+}
+
+// findBenchToCompare will find the last benchmark run on the main branch that is not itself
+function findBenchToCompareIdx(benchRuns) {
+    // Determine what benchmark run we are based on the browser's URL
+    // This is likely the best way to do this without embedding a bunch of data into our profile.js file
+    // or our profile.json file, which although tempting, is not backwards compatible
+    const path = window.location.pathname;
+    const parts = path.split("/");
```
oflatt, line 72: is there not a nice javascript function to get the end of the URL?

infra/nightly-resources/index.js
```diff
@@ -1,46 +1,202 @@
...
+    // have a blank last element, do a quick check and adjust the index accordingly
+    const idx = path[path.length - 1] === "/" ? parts.length - 2 : parts.length - 1;
+    
+    const [date, _, branch, commit] = parts[idx].split("%3A");
+    for (let i = benchRuns.length-1; i >= 0; i--) {
+        const run = benchRuns[i];
+        if (run.branch === "main") {
+            // If we are comparing a run on a main branch, to previous main branch we need to make sure
+            // it is not the same branch.
+            // I did mean `==` here, not `===`. `curComparison.date` is an int, and `date` is a string
```
oflatt, line 86: I think we should use `===` but cast it or convert it to an int explicitly

infra/nightly-resources/index.js
```diff
@@ -1,46 +1,202 @@
...
-        parsed.push(toParse);
-      });
-
-  parsed.sort((l, r) => {
-      if (l.name < r.name) { return -1; }
-      if (l.name > r.name) { return 1; }
-      return 0;
-  });
-  container.innerHTML = ConvertJsonToTable(parsed);
+    throw new Error("Didn't find a candidate benchmark");
```
oflatt, line 96: Shouldn't this say "couldn't find a run from main" or something?

oflatt (changes requested):
Great work! Small things

infra/nightly-resources/index.js
ryan-berger: Nope, `pathname` is all I could find on the interwebs. Weird huh?

infra/nightly-resources/index.js
ryan-berger: That sounds like a good idea to me. I think this function needs a lot more documentation as well, so I'll add some in

infra/nightly-resources/index.js
ryan-berger: Yeah, we could explicitly cast, but I'm not 100% sure I see the need here. Under the hood JS is doing the same thing, and I think that `==` is just as explicit as me writing something like:
```javascript
curComparision.date === +date // this one is a slight hack, but I like the look of it for non-negative numbers
```
or
```javascript
curComparision.date === parseInt(date, 10)
```

oflatt: Yeah I just hate `==` because idk wtf it does. whereas I can look at parseInt and understand.
But this is an optional fix

infra/nightly-resources/index.js
oflatt: dang

infra/nightly-resources/index.js
```diff
@@ -1,46 +1,194 @@
...
-  // aggregate benchmark runs into a list by their "benchmark" key
-  let groupedBy = {};
-  data.forEach((obj) => {
-    if(!groupedBy[obj.benchmark]) {
-      groupedBy[obj.benchmark] = [];
+async function getPreviousRuns() {
+    const req = await fetch("https://nightly.cs.washington.edu/reports-json/eggcc/");
+    const files = await req.json();
+    
+    // map allLinks into an objects of the shape:
```
oflatt, line 5: what is `allLinks`? stale comment?

infra/nightly-resources/index.js
```diff
@@ -1,46 +1,194 @@
...
+    // {branch: <git branch:string>, commit: <git commit:string>, date: <unix timestamp:int>, url: <absolute url to nightly page:string>}
+    const comparisons = [];
+    for (let i = 1; i < files.length; i++) {
+        // file name is of the format <date>:"nightly":<branch>:<commit>
+        const [date, _, branch, commit] = files[i].name.split(":");
```
oflatt, line 10: Unfortunate that we have to parse the file name but oh well

infra/nightly-resources/index.js
```diff
@@ -1,46 +1,194 @@
...
+        comparisons.push(run);
+    }
+
+    comparisons.sort((l, r) => {
+        if (l.date < r.date) {return -1; }
+        if (l.date > r.date) {return 1; }
+        return 0;
+    });
+
+    return comparisons;
```
oflatt, line 32: Are we literally including every run ever in this list? Won't the dropdown eventually become super long?

infra/nightly-resources/index.js
```diff
@@ -1,46 +1,194 @@
...
+}
+
+async function buildNightlyDropdown(element, previousRuns, initialIdx) {
+    const select = document.getElementById(element);
+    
+    const formatRun = (run) => `${run.branch} - ${run.commit}`
```
oflatt, line 38: including the date would be quite helpful

infra/nightly-resources/index.js
```diff
@@ -1,46 +1,194 @@
...
+
+        const sign = entry[key] < prevEntry[key] ? "-" : "+"
+        entry[key] = `${entry[key]} (${sign}${diff})`
+    })
+    return entry;
+}
+
+async function loadBenchmarks(compareTo) {
+    // data should be in the format of Array<{runMethod: String, benchmark: String, total_dyn_inst: Int, hyperfine: Array<{results: **hyperfine `--json` results**}>}>
+    console.log(data);
```
oflatt, line 143: where is `data` bound? not seeing it

oflatt (changes requested):
This is really great! Had a couple questions, but this is almost ready to merge

infra/nightly-resources/index.js
ryan-berger: I guess so, but can't do much when we're working with a file server, not a data-driven API.

infra/nightly-resources/index.js
ryan-berger: Yes, and yes. It does sort by date descending, and we could limit it to 30 or something?

infra/nightly-resources/index.js
ryan-berger: :+1:

infra/nightly-resources/index.js
ryan-berger: `data` is bound to a global that sits in `profile.js` which is an autogenerated file. Here's an example: https://nightly.cs.washington.edu/reports/eggcc/1700941844%3Anightly%3Anightly-comparison%3Ae51adfee2f/data/profile.js

and is generated here: https://github.com/egraphs-good/eggcc/blob/bd8e1f8168e8ae0ed6687dbd58b7e67e597c6406/infra/nightly.sh#L58

infra/nightly-resources/index.js
ryan-berger: Nice catch, I called it `allLinks` before I had a JSON API to work with, now it is `files`. I've updated the docs accordingly

infra/nightly-resources/index.js
oflatt: Sure!

infra/nightly-resources/index.js
oflatt: Ohhh right thanks

oflatt: It seems weird that we scrape the other one we compare to but we have a global for this one. Shouldn't we get the data the same way for both?

ryan-berger: Good question. I think historically we were just making it easier on ourselves so we didn't have to deal with CORS and call `fetch` but now I think it would probably be best if we unified the way we do it. I'll add that to the PR, it isn't a lot of work

infra/nightly-resources/index.js
```diff
@@ -1,46 +1,186 @@
...
+    })
+    return entry;
+}
+
+async function loadBenchmarks(compareTo) {
+    const currentRun = await getBench("../");
+    let previousRun = undefined;
+    try {
+        previousRun = await getBench(compareTo.url);
+    } catch (e) {}
```
oflatt, line 150: Probably should at least print this to the console or something

oflatt (approved):
Great work! I haven't seen any examples where we actually get a diff, so I assume you have tested this.

PR #186, opened 2023-11-15
Change PassThroughArgumentsHelper to VecOperandBase

clyben:
It should reduce the number of VecOperand generated and reduce the VecOperand-get table.


oflatt (approved):
Good change

PR #187, opened 2023-11-16
RVSDG logic

rtjoa:
Learn and propagate "situational" predicates over operands.

So far, the only interesting situation we consider is the last iteration of a Theta, in which we know the predicate must be false. This lets us to rewrite `print res` to `print 8` in the following program, addressing #189:

```
i = 1
do while i < 10
  if i < 10
    res = 7
  else
    res = 8
  i++
print res
```

There are a lot of straightforward ways to extend this proof-of-concept - the standard interval analysis rules, situational constant folding (so far, we only fold situationally-constant Gamma predicates, as is necessary for the above), and adding more situations (e.g. to encode loop invariants; branches of a Gamma).


src/rvsdg/egglog_optimizer/rvsdg-logic.egg
```diff
@@ -0,0 +1,128 @@
...
+(set (refers-to-branch (Node (PureOp (Const (BoolT) (const) (Bool false))))) 0)
+(set (refers-to-branch (Node (PureOp (Const (BoolT) (const) (Bool true))))) 1)
+(rule ((= operand (Node (PureOp (Const (IntT) (const) (Num i))))))
+      ((set (refers-to-branch operand) i))
+      :ruleset fast-analyses)
+
+;; TODO: add the standard interval analysis rules over pureop
+
+; In some context, is the predicate constant?
+(rule ((= gamma (Gamma pred inputs outputs))
```
oflatt, line 98: no single-letter variables pls

src/rvsdg/egglog_optimizer/rvsdg-logic.egg
```diff
@@ -0,0 +1,128 @@
...
+       (holds (leq s1 o1 s2       c)))
+      ((holds (leq s1 o1 (always) c)))
+      :ruleset fast-analyses)
+(rule ((= c (Node (PureOp (Const typ cops lit))))
+       (holds (leq s1       c s2 o2)))
+      ((holds (leq (always) c s2 o2)))
+      :ruleset fast-analyses)
+
+; leq transitivity
+(rule ((holds (leq sa a sb b))
```
oflatt, line 66: Do you actually need this rule for this PR? Doesn't transitivity blow up?

src/rvsdg/egglog_optimizer/rvsdg-logic.egg
```diff
@@ -0,0 +1,128 @@
...
+; leq situation weakening. There are more general ways to write this rule, but
+; we probably mainly want the situations to match
+(rule ((holds (leq s o1 (always) o2)))
+      ((holds (leq s o1 s o2)))
+      :ruleset fast-analyses)
+(rule ((holds (leq (always) o1 s o2)))
+      ((holds (leq s o1 s o2 )))
+      :ruleset fast-analyses)
+
+(relation demand-weakening (Predicate Situation))
```
oflatt, line 52: Unused, remove

oflatt (changes requested):
This PR is very cool, however I'm a little concerned about adding all these situation / predicate combos. We need to think carefully about how to bound the number of these. For example, interval analysis is nice because you get exactly one interval per class. We should restrict it more if possible.

src/rvsdg/egglog_optimizer/rvsdg-logic.egg
rtjoa: Fixed!

src/rvsdg/egglog_optimizer/rvsdg-logic.egg
rtjoa: It's not needed. It doesn't blow up in this case since there are only small set of situations that can interrelate.

I commented it out regardless, but see my broader comment on blowup on your main review.

src/rvsdg/egglog_optimizer/rvsdg-logic.egg
rtjoa: Commented this out

rtjoa:
> This PR is very cool, however I'm a little concerned about adding all these situation / predicate combos. We need to think carefully about how to bound the number of these. For example, interval analysis is nice because you get exactly one interval per class. We should restrict it more if possible.

@oflatt Thanks for the review! I agree that these could easily blow up if we are not careful about how they interact. I commented out the parts you suggested removing

I think the main appeal of an analysis like this is that -- similar to eqsat -- we can write a small number of orthogonal rules that interact in useful and surprising ways. Blowup for now, as there are few situations and we don't create terms (while many other optimizations do). I am inclined to explore this more and find what rules are useful before worrying too much about blowup. Now, it's linear, so it's in `fast-analyses`, but if it gets slower we can just move to another ruleset and bound the number of iterations. Note that the example test case only required a single iteration.

src/rvsdg/egglog_optimizer/rvsdg-logic.egg
```diff
@@ -0,0 +1,131 @@
...
+      :ruleset fast-analyses)
+(rule ((= body (Theta pred inputs outputs)))
+      ((might-create-situation body (always)))
+      :ruleset fast-analyses)
+(rule ((= situation (last-iter pred inputs outputs))
+       (= body (Theta pred inputs outputs)))
+      ((might-create-situation body situation))
+      :ruleset fast-analyses)
+
+(relation holds (Predicate))
```
oflatt, line 41: We don't need holds- just use whether it's in the database or not
didn't catch this the first review sorry

oflatt (changes requested):
Nice work, spotted one more cleanup (my bad)

src/rvsdg/egglog_optimizer/rvsdg-logic.egg
rtjoa: I considered this, but then simply constructing a Predicate (e.g. via `let`) would "make it true" - this feels wrong, and prevents us from having other relations over Predicates.

oflatt (approved):
Fair point- this lgtm

PR #192, opened 2023-11-17
Gamma pull in (try again)

yihozhang:
#177


src/rvsdg/egglog_optimizer/fast_analyses.egg
```diff
@@ -0,0 +1,24 @@
+; Note that because the merge function is set-union
```
oflatt, line 1: I think splitting this code into `.egg` and `.rs` is confusing- I would put all of it in `.rs`

src/rvsdg/egglog_optimizer/fast_analyses.rs
```diff
@@ -167,6 +167,101 @@ fn is_pure_rules() -> Vec<String> {
...
 }
 
+// Tracks what Exprs/Operands/Bodies an operand contains in its context,
+// similar to Body-contains-*.
+// A very important difference is that it does not look into the inputs/outputs of Theta
+// and outputs of Gamma. Instead, it only looks at their inputs.
+//
+// Because there are so many operand, the user of this analysis need
+// to explicitly demand annotate operands they are interested in.
+fn operand_contains_rules() -> Vec<String> {
```
oflatt, line 177: As far as I can tell, you rely on these rules saturating then then use negation to run your rule.
That means that if we miss any cases here, it's unsound. What about vecoperand and operandgroup? I don't see rules for them here.

src/rvsdg/egglog_optimizer/mod.rs
```diff
@@ -33,21 +36,31 @@ pub fn rvsdg_egglog_code() -> String {
...
+        (repeat 2
+            (saturate fast-analyses)
+            (run pull-in)
+            (repeat 100 subst shift)
+        )
+
         ; It is sound to not saturate fast-analyses/subst, but we do because
         ; they won't blow up and will help other rules go through.
-        (repeat 5
+        (repeat 7
```
oflatt, line 56: Did you mean to bump the entire schedule from 5 to 7?

src/rvsdg/egglog_optimizer/gamma-pull-in.egg
```diff
@@ -0,0 +1,65 @@
...
+       (= thn-args (arg-used-VecOperandCtx thn))
+       (set-contains els-args i)
+       (set-not-contains thn-args i)
+
+       (= inp-args (arg-used-Operand inp))
+       )
+      (
+        ;; Step 1: build the new inps: remove (Arg i) and 
+        ;; introduce arguments used by inp-args
+        (let inp-args-vec (VI (set->vec inp-args)))
```
oflatt, line 47: This part is very complex- why do we need to do this?
Can't we assume they use all arguments besides `i`, and just use vec-remove to remove the `i`th input?

src/rvsdg/egglog_optimizer/mod.rs
```diff
@@ -33,21 +36,31 @@ pub fn rvsdg_egglog_code() -> String {
...
     ];
     code.join("\n")
 }
 
 pub fn rvsdg_egglog_schedule() -> String {
     "(run-schedule
+        ;; ad-hoc schedule for gamma pull in optimization
+        (repeat 2
+            (saturate fast-analyses)
+            (run pull-in)
```
oflatt, line 50: Add a comment- it's unsound to not saturate fast-analysis before pull-in

src/rvsdg/egglog_optimizer/subst.rs
```diff
@@ -246,7 +246,7 @@ fn functions_modifying_args(
     ] {
         let fname = func_name_fmt.replace("{}", ty);
         res.push(format!(
-            "(function {fname} ({ty} {aux_params_str}) {ty} :unextractable)",
+            "(function {fname} ({ty} {aux_params_str}) {ty} :cost 10000)",
```
oflatt, line 249: did you mean to leave this in?

oflatt (changes requested):
Left a question- I'd really like to simplify this PR with you before we merge, if possible.

oflatt (approved):
I think we've decided to merge this! Last PR before the rewrite

PR #194, opened 2023-11-22
Make SubstVec*-helpers not to create unnecessary VO/VVO/VOCs

clyben:
Last Tuesday in hackathon we found that VecOperand-get table's rules are firing too many times in some cases, which implies too many VOs are created. A possible cause is subst's helper methods are creating VO/VVOs every iteration. This PR changes the helper only creating new VOBases in iterations. 

actually it reduces test time by about 1s on my computer.
image
image

and it makes the (repeat 1000 subst) in mod.rs line 78 able to saturate, though IDK if it's good idea to make it saturate.


oflatt (approved):
looks good

PR #196, opened 2023-11-28
Create egglog files for the (run) and (check) statements

Alex-Fischman:
Small PR that adds infra for adding optimizations.


oflatt (approved):
good job

PR #202, opened 2023-11-29
Unique args switch rewrites

Alex-Fischman:
This PR adds the two switch optimizations that simplify predicates.

It also contains a bunch of refactors, such as adding saturated rulesets and creating a `util.egg` file.

It also makes `DeepCopy` `unextractable`; I believe this is correct but would like to double check.

Currently missing tests.


oflatt (approved):
Great work!

PR #212, opened 2023-12-18
[Tree unique] IR enums and purity analysis

rtjoa:
We add enums describing the IR (`BRIL_OPS`, but all the way), and use them to concisely generate purity analysis rules.

The main goal of these enums is not to save LOC, but rather:
1. Make us nimble over changes in the IR
2. Reduce bugs caused by egglog typos
3. Get better at egglog metaprogramming

Purity analysis rules populate the following relations:
```
(relation ExprIsPure (Expr))
(relation ListExprIsPure (ListExpr))
``` 

## Stack


tree_unique_args/src/ir.rs
```diff
@@ -0,0 +1,223 @@
...
+            Constructor::Loop => "Loop",
+            Constructor::Body => "Body",
+            Constructor::Arg => "Arg",
+            Constructor::Call => "Call",
+            Constructor::Cons => "Cons",
+            Constructor::Nil => "Nil",
+        }
+    }
+
+    pub(crate) fn fields(&self) -> Vec<Field> {
```
oflatt, line 140: I think fields should probably have an underscore or something to help reduce chance of aliasing

oflatt (approved):
Cool infrastructure, I like the the new `Purpose` abstraction.
Hmm, maybe we should raise an error if you don't provide a ruleset? That might fix our common bug that we run into all the time.

tree_unique_args/src/ir.rs
rtjoa: What might it alias with?

rtjoa:
@oflatt Thanks for the review!

> Hmm, maybe we should raise an error if you don't provide a ruleset? That might fix our common bug that we run into all the time.
This would be a bit of work to enforce statically without building some rule struct directly (rather than strings). I think it's not much of a problem, since with `ir.rs`, we only need to write rule/rewrite a few times per optimization, so it's a lot easier to manually check. Also, we no longer have `(run)` in the schedule, so tests will be more robust to catching this.

tree_unique_args/src/ir.rs
oflatt: Other variables in your pattern.
For example, if I use my own `x` inside a pattern as well as use your fields

rtjoa: Ah, I misread it as saying the method `fields` itself would alias 🤦.

Rather than "default var names," I think field names should be thought of as a more human-readable alternative to indices. They allow us to do [this equality check](https://github.com/egraphs-good/eggcc/blob/63894d78dc04e0f9c0a38cc486bbf221507e80b1/tree_unique_args/src/cf_invariant_code_motion.rs#L17) soundly, and yes, generate variable names better than `x0`, `x1`, `x2`.

I am in favor of not including an underscore since users generating rules can easily append disambiguation themselves if necessary.

oflatt: This is a pretty obvious footgun, we should try to mitigate it

The generated code can be a little messier

rtjoa: Discussed on Slack - added a `.var()` method that prepends an underscore

PR #213, opened 2023-12-18
[Tree unique] Substitution rules

rtjoa:
Add rules for the following functions:
```
(function SubstExpr (Expr Expr) Expr)
(function SubstListExpr (ListExpr Expr) ListExpr)
```

## Stack

### DIFF
--- a/tree_unique_args/src/subst.rs
+++ b/tree_unique_args/src/subst.rs
@@ -27,7 +27,7 @@ fn subst_rule_for_ctor(ctor: Constructor) -> String {
 
 pub(crate) fn subst_rules() -> Vec<String> {
     ESort::iter()
-        .map(|sort| "(function Subst* (* Expr) *)".replace("*", sort.name()))
+        .map(|sort| "(function Subst* (* Expr) * :unextractable)".replace("*", sort.name()))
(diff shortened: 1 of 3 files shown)


tree_unique_args/src/subst.rs
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    let sort = ctor.sort().name();
+    let br = "\n         ";
+    format!("(rewrite (Subst{sort} {ctor_pattern} v){br}{substed_ctor}{br}:ruleset always-run)")
+}
+
+pub(crate) fn subst_rules() -> Vec<String> {
+    ESort::iter()
+        .map(|sort| "(function Subst* (* Expr) * :unextractable)".replace("*", sort.name()))
```
oflatt, line 30: use `format!`

oflatt (approved):
Very nice!

tree_unique_args/src/subst.rs
rtjoa: That's what I did at first, but I changed to `.replace` so we don't have to bind `sort.name()`. I think this is still readable, we often write `Subst*` when talking about families of functions anyway

oflatt: Ok

PR #214, opened 2023-12-25
[Tree unique] `DeepCopy`, `NewLoop`, `NewLet`

rtjoa:
Add rules for the following functions:
```
(function DeepCopyExpr (Expr i64) Expr)
(function DeepCopyListExpr (ListExpr i64) ListExpr)
(function NewLoop (i64 Expr Expr) Expr)
(function NewLet (i64 Expr Expr) Expr)
```

## Stack

### DIFF
--- a/tree_unique_args/src/deep_copy.rs
+++ b/tree_unique_args/src/deep_copy.rs
@@ -1,0 +1,117 @@
+use std::iter;
+
+use crate::ir::{Constructor, ESort, Purpose};
+use strum::IntoEnumIterator;
+
+fn deep_copy_rule_for_ctor(ctor: Constructor) -> String {
+    // e.g. "Add x y"
+    let ctor_pattern = iter::once(ctor.name())
+        .chain(ctor.fields().iter().map(|field| field.name))
+        .collect::<Vec<_>>()
+        .join(" ");
+
+    // e.g. ["(DeepCopyExpr x new-id)", "(DeepCopyExpr y new-id)"]
+    let substed_fields = ctor
+        .fields()
+        .iter()
+        .map(|field| match field.purpose {
+            Purpose::CapturingId => "new-inner-id".to_string(),
+            Purpose::Static(_) => field.name.to_string(),
+            Purpose::CapturedExpr => {
+                let var = field.name;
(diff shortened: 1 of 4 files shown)


tree_unique_args/src/deep_copy.rs
```diff
@@ -0,0 +1,93 @@
...
+    ESort::iter()
+        .map(|sort| "(function DeepCopy* (* i64) * :unextractable)".replace("*", sort.name()))
+        .chain(Constructor::iter().map(deep_copy_rule_for_ctor))
+        .collect::<Vec<_>>()
+}
+
+// We use field names as var names, and bind "v" to the value being substituted
+// in, so this test checks we don't overlap/add extra equality constraints
+#[test]
+fn var_names_available() {
```
oflatt, line 47: Nice test- this is an example of things your field names could alias with.

tree_unique_args/src/sugar.egg
```diff
@@ -10,3 +10,13 @@
...
          :ruleset always-run)
+
+(function NewLoop (i64 Expr Expr) Expr)
+(rewrite (NewLoop id in out)
+         (Loop id (DeepCopyExpr in id) (DeepCopyExpr out id))
+         :ruleset always-run)
+
+(function NewLet (i64 Expr Expr) Expr)
+(rewrite (NewLet id in out)
+         (Let id (DeepCopyExpr in id) (DeepCopyExpr out id))
```
oflatt, line 21: Do we need to copy here? A Let doesn't introduce path conditions so maybe not?

oflatt (approved):
Looks great!

tree_unique_args/src/sugar.egg
rtjoa: I just realized we *shouldn't* copy `in`, since it's in the same context, so I updated the rules accordingly. We do still have to copy `out`, since their contained args correspond to the new input), if we want to be able to union args.

(Also, just to be explicit -`NewLet` and `NewBody` are always intended to be called with `(i64-fresh)` as the first argument)

oflatt: Good fix then

PR #216, opened 2023-12-30
[Tree unique] Conditional invariant code motion

rtjoa:
Lift operators through `Switch`

## Examples

The following rewrite occurs in one iter of the ruleset `conditional-invariant-code-motion`:
```
(Switch
    (Num 1)
    (Pair
        (LessThan (Get (Arg id1) 0) (Num 7))
        (LessThan (Get (Arg id1) 1) (Num 7))
    ))
~~>
(LessThan
    (Switch
        (Num 1)
        (Pair
            (Get (Arg id1) 0)
            (Get (Arg id1) 1)
        ))
    (Num 7))
```

The following rewrite occurs in two iters of the ruleset `conditional-invariant-code-motion` (note that here, the inner `Switch` is just another operator we're lifting! we only lift *through* the outer `Switch`) :
```
(Switch
    (Num id1 0)
    (Pair
        (Switch (LessThan (Get (Arg id1) 0) (Num id1 7)) (Cons (Num id1 11) (Nil)))
        (Switch (LessThan (Get (Arg id1) 1) (Num id1 7)) (Cons (Num id1 11) (Nil)))
    )))
~~>
(Switch
    (LessThan
        (Switch
            (Num id1 0)
            (Pair
                (Get (Arg id1) 0)
                (Get (Arg id1) 1)
            ))
        (Num id1 7))
    (Cons (Num id1 11) (Nil)))
```

## Stack


oflatt (approved):
Great! I finally understand this PR.
What about cases where you want to lift an `Add` but both fields vary in all cases? I think we could do it with tuples and an `Apply` operator or by using a `Let`.
Anyways, future work.

PR #217, opened 2024-01-07
Update rust version and add Cargo.lock to version control

kirstenmg:
Rust version needed to be updated due to a versioning issue with the ahash crate (see [this issue](https://github.com/tkaitchuck/aHash/issues/195)). To (maybe) prevent some future issues from version updates, added Cargo.lock to version control.


oflatt (approved):
Seems good

PR #218, opened 2024-01-08
[Tree unique] Add ids to constants

rtjoa:
This allows us to union with constants.


oflatt (approved):
Good catch, we will want this one

PR #224, opened 2024-01-11
[Tree unique] Add `IdOf` and `ArgOf`

rtjoa:
`IdOf` gets a body's capturing id. `ArgOf` creates an arg that can be placed within a body's captured expression.

### DIFF
--- a/tree_unique_args/src/sugar.egg
+++ b/tree_unique_args/src/sugar.egg
@@ -20,3 +20,16 @@
 (rewrite (NewLet id in out)
          (Let id in (DeepCopyExpr out id))
          :ruleset always-run)
+
+(function IdOf (Expr) i64)
+(rule ((IdOf (Let id in out)))
+      ((set (IdOf (Let id in out)) id))
+      :ruleset always-run)
+(rule ((IdOf (Loop id in out)))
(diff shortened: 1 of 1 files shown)


oflatt:
didn't our old version of IdOf from previous versions work on all expressions, so you could get the id of anything inside a particular context?

rtjoa:
This is useful to make some rules easier to write by not needing to have
multiple versions for let and loop. I didn’t know we had a different
version, which sounds like it has different behavior (returns “referencing
id” rather than “capturing id”)?

oflatt:
Ah I see. Maybe rename to capturing-id-of?

yihozhang:
Can we have two loops equivalent to each other with different ids? e.g. (Loop id1 body1) = (Loop id2 body2). I think this could happen when a rule involving fresh!-i64 is fired twice on the same substitution?

rtjoa:
@yihozhang Two loops with different could totally be equal (imagine they union via a constant). I'll change this to a relation

rtjoa:
Converting to draft as I don't see a use for this if it's a relation. Anyone feel free to inform me otherwise

oflatt:
Closing since we don't need it I think

PR #225, opened 2024-01-11
[Tree unique] Function inlining

kirstenmg:


tree_unique_args/src/function_inlining.egg
```diff
@@ -0,0 +1,16 @@
...
+(rule
+    (
+        (= call (Call f arg))
+        (Function f out)
+    )
+    (
+        (let new-id (Id (i64-fresh!)))
+        (union
+            call
+            ; (NewLet new-id arg out)
```
rtjoa, line 12: Why is NewLet commented, did it not work?

rtjoa (approved):
Nice!! Just a quick question

tree_unique_args/src/function_inlining.egg
kirstenmg: Oops I'll use this now, it was because originally I had included my optimization before `sugar.egg` so `NewLet` didn't exist and I didn't realize why yet.

oflatt:
So clean!

PR #226, opened 2024-01-11
[Tree unique] Add `BodyContains`

rtjoa:
Very similar to https://github.com/egraphs-good/eggcc/pull/143.

Adds relations:
```
(relation BodyContainsExpr (Expr Expr))
(relation BodyContainsListExpr (Expr ListExpr))
```

### DIFF
--- a/tree_unique_args/src/body_contains.rs
+++ b/tree_unique_args/src/body_contains.rs
@@ -1,0 +1,84 @@
+use std::iter;
+
+use crate::ir::{Constructor, ESort, Purpose};
+use strum::IntoEnumIterator;
+
+// Builds rules like:
+// (rule ((Let in out))
+//       ((BodyContains (Let in out) out))
+//       :ruleset always-run)
+fn captured_expr_rule_for_ctor(ctor: Constructor) -> Option<String> {
+    let pat = ctor.construct(|field| field.var());
+    let actions = ctor.filter_map_fields(|field| {
+        (field.purpose == Purpose::CapturedExpr)
+            .then(|| format!("(BodyContainsExpr {pat} {e})", e = field.var()))
+    });
+    if actions.is_empty() {
+        None
+    } else {
+        let actions_s = actions.join(" ");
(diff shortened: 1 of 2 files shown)


tree_unique_args/src/body_contains.rs
```diff
@@ -0,0 +1,82 @@
...
+        Some(format!(
+            "(rule ((BodyContains{sort} body {pat})) ({actions_s}) :ruleset always-run)",
+            sort = ctor.sort().name()
+        ))
+    }
+}
+
+pub(crate) fn rules() -> Vec<String> {
+    ESort::iter()
+        .map(|sort| "(relation BodyContains* (Expr *))".replace("*", sort.name()))
```
ezrosent, line 51: nit: why the `replace("*"` instead of `format!("(relation BodyContains{sort} (Expr {sort}))", sort = sort.name())`, lik e in the rest of this file?

tree_unique_args/src/body_contains.rs
```diff
@@ -0,0 +1,82 @@
+use crate::ir::{Constructor, ESort, Purpose};
+use strum::IntoEnumIterator;
+
+// Builds rules like:
```
ezrosent, line 4: nit, optional: comments above functions should have `///` as the prefix and can also have markdown syntax (which helps rustdoc + vscode views of the docs). So this comment could be

```
/// Builds rules like:
/// ```no_run
/// (rule ((Let in out))
///       ((BodyContains (Let in out) out))
///       :ruleset always-run)
/// ```

tree_unique_args/src/body_contains.rs
rtjoa: The honest reason is that so it doesn't wrap across 3 lines

tree_unique_args/src/body_contains.rs
rtjoa: Thanks, updated!

ezrosent (approved):

PR #227, opened 2024-01-11
Type analysis

ajpal:
This doesn't have everything yet, but the diff is already big, so I think this is a reasonable starting point to merge.
Still to do: loops + functions. Will probably need to add a type environment for those


tree_unique_args/src/subst.rs
```diff
@@ -45,6 +45,28 @@ fn var_names_available() {
     }
 }
 
+#[test]
```
oflatt, line 48: Nice test! I suggest the following for the final version
- Add a couple tests that give tuple types
- Make the run_test method check that your typechecker gives the program a type (to ensure you are not missing rules)
- Write a rule that panics if something gets two different types

oflatt:
I agree this is a good stopping point. Made follow-up tasks: #270 and #271

oflatt:
This closes #238

tree_unique_args/src/type_analysis.egg
```diff
@@ -0,0 +1,231 @@
...
+
+; Effectful Ops
+(rule ((HasTypeDemand (Print e)))
+      ((HasType (Print e) (UnitT)))
+      :ruleset type-analysis)
+(rule ((HasTypeDemand (Write addr val)))
+      ((HasType (Write addr val) (UnitT)))
+      :ruleset type-analysis)
+(rule ((HasTypeDemand (Read addr)))
+      ((HasType (Read addr) (IntT))) ; all reads are ints
```
oflatt, line 169: In Bril reads are not all integers, so we'll have to add type annotations to reads.
I would say for now lets leave read and write out, made an issue

tree_unique_args/src/type_analysis.egg
```diff
@@ -0,0 +1,231 @@
...
+      :ruleset type-analysis)
+
+; Switch 
+; peel off a branch and demand type
+(rule ((HasTypeDemand (Switch cond (Cons branch rest))))
+      (
+       (HasTypeDemand branch)
+       (HasTypeDemand (Switch cond rest))
```
oflatt, line 177: This is a cool solution but I think it has a problem
This creates a `Switch` which has the wrong number of branches compared to its condition. I think we don't want to do that (it would create a term that doesn't typecheck)

Instead, I propose doing a `ListHasUniformTypeDemand` and a corresponding `ListHasUniformType` relation. `ListHasUniformType` would be set to a type if all the elements in the list have the same type, which is what you need to typecheck a `Switch`.

tree_unique_args/src/type_analysis.egg
```diff
@@ -0,0 +1,231 @@
...
+(rule (
+       (HasTypeDemand (Switch cond (Cons branch rest)))
+       (HasType (Switch cond rest) ty)
+       (HasType branch ty)
+      )
+      ((HasType (Switch cond (Cons branch rest)) ty))
+      :ruleset type-analysis)
+
+; Sequencing
+(rule ((HasTypeDemand (All ord (Cons hd tl))))
```
oflatt, line 203: Optional: I would put all the demand rules together in the file, then all the typechecking rules.

tree_unique_args/src/type_analysis.egg
```diff
@@ -0,0 +1,231 @@
...
+      :ruleset type-analysis)
+; base case- demand the type of the condition
+(rule ((HasTypeDemand (Switch cond Nil)))
+      ((HasTypeDemand cond))
+      :ruleset type-analysis)
+
+; base case- single branch switch has type of branch
+(rule (
+       (HasTypeDemand (Switch cond (Cons branch (Nil))))
+       (HasType cond BoolT)
```
oflatt, line 188: I think this IR supports integer conditions. I guess it should typecheck for booleans and integers? For booleans we sure ensure two branches

oflatt (changes requested):
Found some small problems, but this is on the right track!

oflatt (approved):
Looks great!

PR #229, opened 2024-01-11
bool rules, constant condition, switch pull in beneath

ajpal:
```
1. !(a or b) ==> !a and !b
2. !(a and b) ==> !a or !b
3. !! a ==> a
4. if true A B ==> A
5. if false A B ==> B
6. (if E then S1 else S2); S3 ==> if E then S1;S3 else S2;S3
```


oflatt (approved):
Looks good to me! Nice

yihozhang:
Note: This PR does seem to handle cases like
```
var x = 0;
if b {
  x = 1;
} else {
  x = -1;
}
return f(x)
```
whose corresponding IR form is
```
f(if b then 1 else -1)
```
since there are no side effects. So this PR actually implements a different optimization than the switch pull in beneath from the milestone encoding?

PR #245, opened 2024-01-12
Add util.egg

clyben:
Add util.egg, which contains ListExpr-length and ListExpr-ith function.

### DIFF
--- a/tree_unique_args/src/util.egg
+++ b/tree_unique_args/src/util.egg
@@ -1,0 +1,13 @@
+(function ListExpr-length (ListExpr) i64)
+(function ListExpr-ith (ListExpr i64) Expr :unextractable)
+(function ListExpr-surfix (ListExpr i64) ListExpr :unextractable)
+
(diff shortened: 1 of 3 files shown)


tree_unique_args/src/util.egg
```diff
@@ -0,0 +1,13 @@
+(function ListExpr-length (ListExpr) i64)
+(function ListExpr-ith (ListExpr i64) Expr :unextractable)
+(function ListExpr-surfix (ListExpr i64) ListExpr :unextractable)
```
oflatt, line 3: Typo: I think you meant "suffix"

oflatt (changes requested):
Looks good! I think you have a typo

tree_unique_args/src/util.egg
clyben: ohh....(

oflatt (approved):
Thanks!
For future reference, after you fix up a PR don't forget to re-request a review.

PR #248, opened 2024-01-12
[Tree unique] Simplify loops to not have inputs

oflatt:
It turns out that loop inputs are unnecessary, and we can just use `Let`. So this rule removes inputs from loops. For the first iteration, the argument in the loop refer to the argument outside the loop.

For example, the following loop:
```
(Loop (Id 1)
  (Add (Num 1) (Num 2))
   some_loop_body)
```

Can be expressed as 
```
(Let (Id 2)
  (Add (Num 1) (Num 2))
    (Loop (Id 1) some_loop_body))
```

In general, the following rule holds:
```
(rule ((Loop inputs outputs))
      ((let new-id (i64-fresh!))
       (union (Loop inputs outputs)
              (NewLet new-id inputs (Loop (Arg new-id) outputs)))))
```


Alex-Fischman (commented):
I think that this should not be merged as-is for software engineering reasons; we shouldn't change the schema out from under people if we can avoid it.

I'm also not convinced that this is a good idea, even though in general I'm all for making changes like this. I just don't see how this change will make our optimizations simpler.

oflatt:
Converting this to a draft- I think we should consider it in the future but not implement it right now.

oflatt:
Closing, I think it's more natural for loops to have inputs like other operators

PR #250, opened 2024-01-15
[Tree unique] WIP Loop unrolling 

oflatt:
This PR adds a work-in-progress loop unrolling rule.

The rule is inefficient because it uses multiple layers of deep copying. I think this could be made more efficient by only having a top-level deep copy- however, that would add invalid intermediate terms to the database. We should decide how to resolve this issue.

On the other hand, maybe this does not matter. Doing loop unrolling in this IR is quite inefficient anyways, since each unrolling doubles the size of the sub-egraph.

### DIFF
--- a/tree_unique_args/src/sugar.egg
+++ b/tree_unique_args/src/sugar.egg
@@ -26,10 +26,4 @@
          (Let id in (DeepCopyExpr out id))
          :ruleset always-run)
 
-(function FreshLet (Expr Expr) Expr :unextractable)
-(rule ((FreshLet in out))
-      ((let new-id (Id (i64-fresh!)))
-       (Let new-id in (DeepCopyExpr out new-id)))
-       :ruleset always-run)
-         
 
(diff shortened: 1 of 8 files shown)


tree_unique_args/src/sugar.egg
```diff
@@ -1,8 +1,13 @@
 ; Functions useful for writing tests and actions, but should not be matched on
 (function Pair (Expr Expr) ListExpr)
-(rewrite (Pair a b)
-         (Cons a (Cons b (Nil)))
-         :ruleset always-run)
+(birewrite (Pair a b)
```
oflatt, line 3: Changed to birewrite so we can query pairs

tree_unique_args/src/util.egg
```diff
@@ -11,3 +11,20 @@
 (rule ((= (ListExpr-suffix list n) (Nil)))
     ((set (ListExpr-length list) n)) :ruleset always-run)
 
+;; given a start i and an end j,
+;; constructs a list
+;; [(Get (Arg (Id 0)) i), (Get (Arg (Id 0)) i+1), ..., (Get (Arg (Id 0)) j-1)]
+(function SubArgList (i64 i64) Expr :unextractable)
```
oflatt, line 17: Useful for getting sub-tuples

oflatt:
Closing this because of deep copy issues. Trying again on new schema

PR #258, opened 2024-01-17
Parse egglog into Expr

### LINKED ISSUE #249: [Utility] Make interpreter work on egglog s-expressions
Currently, the interpreter for the tree-unique IR works on a rust enum `Expr`, not the s-expressions we type to write egglog tests. Unfortunately this means we can't run the interpreter on our tests, which would help a lot.
We should write a parser that converts to `Expr` so we can do so

Alex-Fischman:
- Update `Expr` to match the schema.
- Implement `FromStr` for `Expr`
- Add a test for the `FromStr` implementation

Resolves #249.

### DIFF
--- a/tree_unique_args/tests/interpreter.rs
+++ b/tree_unique_args/tests/interpreter.rs
@@ -2,16 +2,16 @@
 
 use std::collections::HashMap;
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum Order {
     Parallel,
     Sequential,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct Id(i64);
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum Expr {
     Num(i64),
     Boolean(bool),
@@ -410,8 +410,20 @@ fn test_interpreter_fib_using_memory() {
 impl std::str::FromStr for Expr {
     type Err = String;
     fn from_str(s: &str) -> Result<Expr, String> {
-        fn list_expr_to_vec(_e: &egglog::ast::Expr) -> Result<Vec<Expr>, String> {
-            todo!("need to parse a cons list")
+        fn list_expr_to_vec(e: &egglog::ast::Expr) -> Result<Vec<Expr>, String> {
+            if let egglog::ast::Expr::Call(f, xs) = e {
+                match (f.as_str(), xs.as_slice()) {
+                    ("Nil", []) => return Ok(Vec::new()),
+                    ("Cons", [head, tail]) => {
+                        let head = egglog_expr_to_expr(head)?;
+                        let mut tail = list_expr_to_vec(tail)?;
+                        tail.insert(0, head);
+                        return Ok(tail);
+                    }
+                    _ => {}
+                }
+            }
+            Err(String::from("invalid ListExpr"))
         }
         fn egglog_expr_to_id(e: &egglog::ast::Expr) -> Result<Id, String> {
             if let egglog::ast::Expr::Call(f, xs) = e {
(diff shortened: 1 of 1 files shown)


tree_unique_args/tests/interpreter.rs
```diff
@@ -182,6 +193,15 @@ pub fn interpret(e: &Expr, arg: &Option<Value>, vm: &mut VirtualMachine) -> Valu
             };
             Value::Num(n1 - n2)
         }
+        Expr::Mul(e1, e2) => {
+            let Value::Num(n1) = interpret(e1, arg, vm) else {
+                panic!("mul")
+            };
+            let Value::Num(n2) = interpret(e2, arg, vm) else {
```
oflatt, line 200: I think you can leave off the else cases since they should be typechecked. I think rust just raises an error in the else by default

tree_unique_args/tests/interpreter.rs
```diff
@@ -147,6 +155,9 @@ pub fn typecheck(e: &Expr, arg_ty: &Option<Type>) -> Result<Type, TypeError> {
             typecheck(output, &Some(input_ty.clone()))
         }
         Expr::Arg(_) => arg_ty.clone().ok_or(TypeError::NoArg(e.clone())),
+        // TODO: add an environment for functions so we can typecheck function calls correctly
```
oflatt, line 158: sounds like this is a good next task, could you add an issue for it?

tree_unique_args/tests/interpreter.rs
```diff
@@ -361,3 +406,153 @@ fn test_interpreter_fib_using_memory() {
     assert_eq!(vm.mem[&(nth as usize)], Value::Num(fib_nth));
     assert!(!vm.mem.contains_key(&(nth as usize + 1)));
 }
+
+impl std::str::FromStr for Expr {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Expr, String> {
+        fn list_expr_to_vec(e: &egglog::ast::Expr) -> Result<Vec<Expr>, String> {
```
oflatt, line 413: I would prefer an error like the `EggCCError` we have in the milestone code

tree_unique_args/tests/interpreter.rs
```diff
@@ -361,3 +406,153 @@ fn test_interpreter_fib_using_memory() {
...
+                {
+                    return Ok(Id(*int));
+                }
+            }
+            Err(String::from("invalid Id"))
+        }
+        fn egglog_expr_to_expr(e: &egglog::ast::Expr) -> Result<Expr, String> {
+            match e {
+                egglog::ast::Expr::Lit(_) => Err(String::from("literals must be wrapped")),
+                egglog::ast::Expr::Var(s) => Err(format!("unknown variable {:?}", s.as_str())),
```
oflatt, line 441: Maybe a more descriptive error would say "expected a grounded term, got a term with a variable in it: "

tree_unique_args/tests/interpreter.rs
```diff
@@ -257,6 +301,7 @@ pub fn interpret(e: &Expr, arg: &Option<Value>, vm: &mut VirtualMachine) -> Valu
             let Some(v) = arg else { panic!("arg") };
             v.clone()
         }
+        Expr::Function(_, _) | Expr::Call(_, _) => todo!("interpret functions and calls"),
```
oflatt, line 304: Another good issue to create. I think what would be best is to make a top level `(Program (Cons func_1 ...))` that we could interpret with some inputs.

oflatt (changes requested):
Nicely done, excited to get this working with our tests so we can run things. Had some small pieces of feedback

tree_unique_args/tests/interpreter.rs
Alex-Fischman: a) Ryan put those there
b) Pretty sure it doesn't compile without them

oflatt (approved):
Great!

PR #259, opened 2024-01-17
Fix several bugs in structured conversion

oflatt:
I've re-visited the Bril `Program` to `StructuredProgram` conversion code because I would like to use it for the tree unique schema.

There were two bugs to fix:
- Structured conversion considered `If` to be a scope in some places but not others
- Breaking out of a block within a loop was incorrect and broke out too far

### DIFF
--- a/src/cfg/structured.rs
+++ b/src/cfg/structured.rs
@@ -1,6 +1,6 @@
 use std::fmt::Display;
 
-use super::BasicBlock;
+use super::{function_to_cfg, BasicBlock};
 use bril_rs::{Argument, Code, EffectOps, Function, Instruction, Program, Type};
 
 #[derive(Debug, PartialEq, Clone)]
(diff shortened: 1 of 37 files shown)


src/cfg/structured.rs
```diff
@@ -188,16 +191,36 @@ impl StructuredBlock {
...
                         pos: None,
                     }));
+
+                // convert the then block
                 builder.resulting_code.push(Code::Label {
                     label: then_name,
                     pos: None,
                 });
                 then_block.to_code(builder);
+                // jump to the end
```
oflatt, line 201: this was another bug where the then branch was falling through

src/cfg/to_structured.rs
```diff
@@ -260,12 +294,13 @@ impl<'a> StructuredCfgBuilder<'a> {
...
 
+    /// merge nodes have multiple incoming
+    /// edges
     fn is_merge_node(&self, node: NodeIndex) -> bool {
         self.cfg
             .graph
             .neighbors_directed(node, petgraph::Direction::Incoming)
-            .take(1)
-            .next()
+            .nth(1)
```
oflatt, line 303: another bug where merge nodes were not being detected correctly

Alex-Fischman (approved):
LGTM

PR #262, opened 2024-01-19
[Tree unique] Add `{Expr,ListExpr}IsValid` & update existing optimizations

rtjoa:
Adds the following relations, which indicate that we expect some IR to be well-formed (typecheck, proper use of referencing and capturing ids).
```
(relation ExprIsValid (Expr))
(relation ListExprIsValid (ListExpr))
```

All optimizations that add to the e-graph should first match on these relations, to make sure we optimize code we care about, and not intermediary terms that still need to be substed or copied. This has the following benefits:
- Preserve invariants - it prevents provenance of an Expr term from having semantics, which makes [soundness of our capturing id invariants](https://gist.github.com/rtjoa/a28a0dd7b16da2f4a5a414ad8c4c7127/) easier.
- Performance - when we only optimize `IsValid` expressions, we avoid wasting time on intermediary to-be-copied-or-substed expressions, which could be a source of blowup.

This PR also ports existing optimizations to use these relations.

### DIFF
--- a/tree_unique_args/src/conditional_invariant_code_motion.rs
+++ b/tree_unique_args/src/conditional_invariant_code_motion.rs
@@ -44,7 +44,7 @@ fn rules_for_ctor(ctor: Constructor) -> Option<String> {
                          :ruleset always-run)
 
                 ; Lift {ctor_name} when only {varying_field_name} varies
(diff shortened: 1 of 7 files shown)


oflatt (approved):
Great!

PR #263, opened 2024-01-19
[Tree unique] Loop strength reduction

kirstenmg:
Implement the optimization loop strength reduction.
Note that this does not work for a case covered in previous iterations of the optimization: where the constant being multiplied is a Num, rather than an invariant input. This is because the constant needs to be added to an input of the loop, and so needs to use the ID of the loop's outer context. For invariant inputs, we just use the value of the invariant from the loop input. However, for a Num, we need to change its Id to use it in the input, since we only matched on a num from within the Loop body. To get the id of the Loop's outer context, we need RefIdOf.


kirstenmg:
This is dependent on [referencing id analysis](https://github.com/egraphs-good/eggcc/pull/265), but since the id analysis branch is on my own fork I don't think I can configure GitHub to show just the diff between this branch and the id analysis branch in this PR.

yihozhang:
Turning this into draft for now. Let me know when it is ready for review!

tree_unique_args/src/schedule.egg
```diff
@@ -3,11 +3,12 @@
 (run-schedule
   (repeat 6
     (saturate always-run)
-    (saturate error-checking) ; In the future, this will be "debug mode" only
+    (saturate error-checking) ; In the future, this will be "release mode" only
```
yihozhang, line 6: What does this mean? Is it because error-checking is too slow for rustc's debug mode? I would think we need to do error checking if we are in *eggcc*'s debug mode. A small clarification here would be helpful.

tree_unique_args/src/loop_strength_reduction.egg
```diff
@@ -0,0 +1,111 @@
...
+;; for i = 0 to n:
+;;     a = d
+;;     d += c
+
+; Finds invariants/constants within a loop.
+; Columns: loop; value of invariant in inputs; value of invariant in outputs
+; TODO: once everything else is working, change this to just use an ID to ref the loop
+(relation Inv (Expr Expr Expr))
+(rule (
+    (= loop (Loop id (All (Parallel) inputs) (All (Sequential) (Pair pred (All (Parallel) outputs)))))
```
yihozhang, line 24: My understanding is that a `Pair` can be desugared into `Cons/Nil`, but not the other direction, so having `Pair` in a rule's trigger may make the rule incomplete.

tree_unique_args/src/loop_strength_reduction.egg
```diff
@@ -0,0 +1,111 @@
...
+; belongs within the context of the outputs, and therefore relies on RefIdOf.
+; TODO: handle this and add corresponding test case
+; This note I copied from the previous encoding (unique args) is actually incorrect:
+    ;; Note: with the milestone encoding, we had to use
+    ;; Body-contains-Operand here. Here, we already know
+    ;; that c is within loop because it has the correct id. 
+(rule (
+    (= c (Num id n))
+    (= loop (Loop id inputs predoutputs))
+    (ExprHasRefId loop outer-id))
```
yihozhang, line 39: Indeed, with newly merged #291, you can get the id of outer region of `loop` by inspecting the id to inputs.

tree_unique_args/src/loop_strength_reduction.egg
```diff
@@ -0,0 +1,111 @@
...
+    ((Inv loop (Num outer-id n) c)) :ruleset always-run)
+
+(rule
+    (
+        ;; Find loop
+        (= old-loop (
+            Loop
+            id
+            (All (Parallel) input-list)
```
yihozhang, line 48: Does input have to be `Parallel` and outputs have to be `Sequential`?

tree_unique_args/src/loop_strength_reduction.egg
```diff
@@ -0,0 +1,111 @@
...
+                    )
+                )))
+            )
+        (ExprIsValid old-loop)
+
+        ; Find loop variable (argument that gets incremented with a number)
+        (= (ListExpr-ith output-list i) (Add (Get (Arg id) i) (Num id loop-incr)))
+
+        ; Find invariant where input is same as output
+        ; TODO: no longer works for multiplication by Num, since changing the id of the Num
```
yihozhang, line 64: This comment is obsolete?

tree_unique_args/src/loop_strength_reduction.egg
```diff
@@ -0,0 +1,111 @@
...
+                        pred
+                        new-outputs))))
+
+        ; We can just union within this new loop now
+        ; NOTE: milestone encoding used can-subst-beneath for this
+        (union (DeepCopyExpr old-exp new-id) (Get (Arg new-id) n))
+
+        ; NOTE: with single args, I believe it's ok to just union this
+        ; without adjusting for the number of inputs/outputs
+        (union old-loop new-loop)
```
yihozhang, line 108: I'm actually not sure here- `old-loop` produces a tuple of `n` values and `new-loop` produces a tuple of `n+1` tuples, so they are not supposed to be equivalent?

yihozhang (changes requested):
This is awesome! Thanks for making it. I left some questions.

tree_unique_args/src/loop_strength_reduction.egg
```diff
@@ -0,0 +1,111 @@
...
+;;     a = d
+;;     d += c
+
+; Finds invariants/constants within a loop.
+; Columns: loop; value of invariant in inputs; value of invariant in outputs
+; TODO: once everything else is working, change this to just use an ID to ref the loop
+(relation Inv (Expr Expr Expr))
+(rule (
+    (= loop (Loop id (All (Parallel) inputs) (All (Sequential) (Pair pred (All (Parallel) outputs)))))
+    (= (ListExpr-ith outputs i) (Get (Arg id) i)))
```
yihozhang, line 25: Specific to the way the optimization is currently written, we also need to ensure that `(ListExpr-ith inputs i)` is pure, since `(ListExpr-ith inputs i)` will get duplicated later.

I'm also slightly worried that if `(ListExpr-ith inputs i)` is a very expensive (but pure) operation, this optimization actually makes the performance worse and, more importantly, no optimization rule I can think of can recover the performance (no rule can remove `(Get (Args id) i)` from the loop body).

tree_unique_args/src/loop_strength_reduction.egg
oflatt: I think we changed it so that we have the other direction too

oflatt:
Really great work, but we are focusing on the new schema. Close this for now?

PR #265, opened 2024-01-20
[Tree unique] Referencing Id Analysis

kirstenmg:
Creates relations `ExprHasRefId`, `ListExprHasRefId` to find the referencing id of an expression (the id of the context containing the expression).

### DIFF
--- a/tree_unique_args/src/id_analysis.rs
+++ b/tree_unique_args/src/id_analysis.rs
@@ -9,11 +9,15 @@ fn id_analysis_rules_for_ctor(ctor: Constructor) -> String {
         let field_sort = field.sort().name();
         match field.purpose {
             Purpose::Static(_) | Purpose::CapturedExpr | Purpose::CapturingId => None,
+
+            // Base case: constructor has referencing id specified as a field
             Purpose::ReferencingId => Some(format!(
                 "(rule ({pat})
                        (({sort}HasRefId {pat} {field_var}))
                        :ruleset always-run)"
             )),
+
+            // Constructor has referencing id of its subexpr fields
             Purpose::SubExpr | Purpose::SubListExpr => Some(format!(
                 "(rule ({pat} ({field_sort}HasRefId {field_var} ref-id))
                        (({sort}HasRefId {pat} ref-id))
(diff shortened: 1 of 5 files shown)


kirstenmg:
After merging with main, the following tests fail because the ref ids don't match:
`    simple::test_simple_bool_rules
    switch_rewrites::switch_pull_in_below
    switch_rewrites::test_constant_condition`
I will figure this out then re-open the PR.

tree_unique_args/src/schedule.egg
```diff
@@ -3,6 +3,7 @@
 (run-schedule
   (repeat 5
     (saturate always-run)
+    (saturate error-checking) ; In the future, this will be "release mode" only
```
oflatt, line 6: did you mean "debug mode" only?
I think it's fine to run these in release unless we think they are slow

tree_unique_args/src/id_analysis.rs
```diff
@@ -0,0 +1,112 @@
...
+(rule ((ListExprHasRefId x id1)
+       (ListExprHasRefId x id2)
+       (!= id1 id2))
+      ((panic \"Ref ids don't match\"))
+      :ruleset error-checking)
+    "
+    .to_string()];
+
+    ESort::iter()
+        .map(|sort| "(relation *HasRefId (* IdSort))".replace('*', sort.name()))
```
oflatt, line 48: Here you map over the sorts, but above you implement for `Expr` and `ListExpr` manually.
Instead you could add your error checking rule in this templated string

oflatt (changes requested):
Really great job on implementing this, testing, and fixing other tests.
Had a couple small comments

tree_unique_args/src/schedule.egg
kirstenmg: Yes, pretty sure this should be debug mode only (if other error-checking rules end up slow)

oflatt (approved):
Great work! Before we merge, could you add a test that it catches a couple bad cases?
You could add a couple
```
#[test]
#[should_panic]
```

PR #267, opened 2024-01-22
Interval Analysis for Tree Unique

### LINKED ISSUE #238: [Tree unique] Typechecking terms without loops

ajpal:
Basically the same logic as https://github.com/egraphs-good/eggcc/pull/138, but much simpler under new schema

Future work
- generalize rules to more branches
- add more interval rules for the math operators. Right now just + and <

Closes #238


tree_unique_args/src/interval_analysis.egg
```diff
@@ -0,0 +1,75 @@
...
+       (= (IntI la ha) (ival a))
+       (= (IntI lb hb) (ival b))
+      )
+      ((set (ival lhs) (IntI (+ la lb) (+ ha hb))))
+      :ruleset interval-analysis)
+
+; Constant condition elimination with interval analysis
+(rule (
+       (= lhs (Switch pred (Cons A (Cons B (Nil)))))
+       (= (BoolI true true) (ival pred))
```
yihozhang, line 57: Need to ensure `pred` is pure? I find that you almost always want to use `ival` in conjunction with `ExprIsPure`, so if this is true, maybe a comment on the definition of `ival` would be helpful.

ezrosent: If `ExprIsPure` then could we replace these two rules with
```
(rule ((ExprIsPure pred) (= (BoolI x x) (ival pred))) ((union pred x)))
```

Then rely on the existing rewrites for Switch to collapse this into `A` or `B`?

ezrosent (changes requested):
I think we need `ExprIsValid` everywhere now?

tree_unique_args/src/interval_analysis.egg
ajpal: This diff removes the constant condition rules for Switch since they're a special case of the interval analysis.

ezrosent: Sorry, I was more wondering why in general we wouldn't union expressions with only one possible value per the interval with the constant itself, then keep the various constant-folding rules as-is. 

I appreciate that writing all of these various rules in terms of the interval analysis gives us more generality though.

ezrosent (approved):

tree_unique_args/src/interval_analysis.egg
ajpal: ohh got it, yeah that's better. Fixed

PR #268, opened 2024-01-22
[TreeUnique] IVT

ezrosent:
This is my attempt at #244 

The rules are a lot simpler than in the milestone encoding. That's due to the encoding itself being easier to write rules for, and also because the rule itself only matches loops whose body is "just a switch." I think we want to have rules that push computations _inside_ switches in order to expose opportunities to IVT (we already have some of these).

Marking this as a draft as I'm not sure about the best way to re-assign expressions' ids (e.g. within `Num`s) to ids from the outer scope. The test gives an example (see the `TODO`) of a case where the `Num`s retain the id from the old loop. I figure this is something that's come up before and I'd appreciate guidance on on the cleanest way to do it.

### DIFF
--- a/tree_unique_args/src/ivt.egg
+++ b/tree_unique_args/src/ivt.egg
@@ -32,7 +32,7 @@
 (rule ((= loop (Loop id in out))
        (ExprHasRefId loop outer-id)
        (ExprIsValid loop)
-       (= out (All ord (Pair pred switch)))
+       (= out (All ord (Cons pred (Cons switch (Nil)))))
        (= switch (Switch pred branches))
        (ExprIsPure pred))
       ((LiftSwitch switch in outer-id)) :ruleset ivt)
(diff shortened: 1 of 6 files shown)


tree_unique_args/src/ivt.egg
```diff
@@ -0,0 +1,52 @@
...
+         (DeepCopyExpr (SubstExpr inner subst) outer-id) :ruleset always-run)
+
+; Create demand
+(rule ((= loop (Loop id in out))
+       (ExprHasRefId loop outer-id)
+       (ExprIsValid loop)
+       (= out (All ord (Cons pred (Cons switch (Nil)))))
+       (= switch (Switch pred branches))
+       (ExprIsPure pred))
+      ((LiftSwitch switch in outer-id)) :ruleset ivt)
```
kirstenmg, line 38: Why do you need to separate lifting the switch from applying the following rule? Could you not lift the switch within the actions for the next rule?

kirstenmg (approved):
This looks good to me, I just have one question for my own understanding but don't think it impacts correctness.

tree_unique_args/src/ivt.egg
ezrosent: The issue here is that I only want to do the `(DeepCopy (Subst ...))` business if I have a potential match for IVT. But once I _do_ have such a match, I have to wait for DeepCopy and Subst to run so I can extract out the `thn` and `els` of the new switch.

The pattern here is to essentially run the LHS for IVT twice, once to kick off the process of lifting the switch, and a second time to create the entire new expression and union it with the loop.

ezrosent:
Thanks for the review! And let me know if I didn't understand your comment correctly; happy to send a follow-up if there are more ways to improve here.

PR #269, opened 2024-01-22
[Tree unique] Loop Invariant Detection

clyben:
This PR implement Loop invariant detection


tree_unique_args/src/ir.rs
```diff
@@ -205,6 +205,17 @@ impl Constructor {
         format!("({without_parens})")
     }
 
+    pub(crate) fn construct_only_fields<F>(&self, f: F) -> String
```
yihozhang, line 208: Do we need this function? It seems this can be implemented in terms of `filter_map_fields`.

tree_unique_args/src/util.rs
```diff
@@ -35,3 +35,38 @@ fn append_test() -> Result<(), egglog::Error> {
 
     crate::run_test(build, check)
 }
+
+#[test]
+fn get_loop_output_ith_test() -> Result<(), egglog::Error> {
```
yihozhang, line 40: Nice test for `get-loop-output-ith`!

yihozhang (commented):
I made an initial pass and I think the rules make sense to me. I will make another pass shortly to see if there are places where we can make the rule-generating code more precise. It would be great to clean up the PR for comments, typos, etc as well. Overall great work!

tree_unique_args/src/ir.rs
clyben: The function was a bit more complex at beginning. I can remove it.

clyben: Done, also collapse the code in rule generation.

tree_unique_args/src/loop_invariant.egg
```diff
@@ -0,0 +1,43 @@
+;; Loop Invariant
+(relation find-inv-Expr (Expr Expr))
```
oflatt, line 2: Could this be a more generic "loop-child" relation that can be re-used by other optimizations?

tree_unique_args/src/loop_invariant.egg
```diff
@@ -0,0 +1,43 @@
...
+(relation find-inv-ListExpr (Expr ListExpr))
+
+;; bool: whether the term in the Expr is an invariant.
+(function is-inv-Expr (Expr Expr) bool :unextractable :merge (or old new))
```
oflatt, line 6: Why have an output bool here?
Why not have a relation instead, and match on the relation?

tree_unique_args/src/loop_invariant.rs
```diff
@@ -0,0 +1,199 @@
...
+
+use crate::ir::{Constructor, Purpose, Sort};
+
+fn find_invariant_rule_for_ctor(ctor: Constructor) -> Option<String> {
+    let br = "\n      ";
+    let ruleset = " :ruleset always-run";
+
+    match ctor {
+        Constructor::Cons | Constructor::Nil | Constructor::Arg | Constructor::UnitExpr => None,
+        Constructor::Call => Some(format!(
```
oflatt, line 12: I think we should split out `find-in-Expr` into it's own analysis file, and rename it to `loop-child`

tree_unique_args/src/loop_invariant.rs
```diff
@@ -0,0 +1,199 @@
...
+        // print, read, write are not invariant
+        // assume Arg as whole is not invariant
+        Constructor::Cons
+        | Constructor::Nil
+        | Constructor::UnitExpr
+        | Constructor::Print
+        | Constructor::Read
+        | Constructor::Write
+        | Constructor::Arg => None,
+        Constructor::Get => Some(format!(
```
oflatt, line 71: Why is get special? It seems like the `_` case would work?

oflatt (changes requested):
Really awesome work, I'm excited about this. Had a few questions

tree_unique_args/src/loop_invariant.egg
clyben: I think the generic one is is-inv-Expr function

tree_unique_args/src/loop_invariant.egg
clyben: The reason is for boundary analysis. In boundary analysis we also want to know if an Expr is not invariant (to find the boundary where an invariant's parent is not invariant), so in default I set all to be false and change when finding the invariant.

tree_unique_args/src/loop_invariant.rs
clyben: Because the Get have a (Purpose::Static(Sort::I64), "i") in fields. If it pass to `_` case it generate a `(set (is-inv-Expr loop expr) true)` which is unwanted. I feel like this `i` is different from the `Purpose::Static` in `Num` and `Bool` but easy solution now is to have special case here.

tree_unique_args/src/loop_invariant.rs
clyben: Is that the same as BodyContains?

clyben: This find-inv relation might be replaced by body-constain?

oflatt: Yeah, you could also do a more general `body-contains`

tree_unique_args/src/loop_invariant.rs
oflatt: Why is it different from `Num` and `Bool`?
I think I still don't quite understand.

clyben: Oh you are right, the match on Get in line71 is not necessary, though the match on Get on line16 is needed.

oflatt:
re-request review when it's ready

clyben:
@oflatt Changed to BodyConstainExprs

clyben:
issue #299 now block it since this PR now use BodyContains

tree_unique_args/src/loop_invariant.egg
```diff
@@ -0,0 +1,28 @@
+;; Loop Invariant
+
+;; bool: whether the term in the Expr is an invariant.
+(function is-inv-Expr (IdSort Expr) bool :unextractable :merge (or old new))
+(function is-inv-ListExpr (IdSort ListExpr) bool :unextractable :merge (or old new))
+
+(relation arg-inv (IdSort i64))
```
oflatt, line 7: Why do we need `arg-inv` and `is-inv-Expr`?
Isn't `(arg-inv id i)` equivalent to `(is-inv-Expr id (Get (Arg id) i))`

clyben: True, it was used before changing to BodyContains for trigering find-invariant, but now it's not needed

clyben: fixed

yihozhang:
As promised I made another pass and the PR at this stage looks very clear and easy to follow. Nice work!

oflatt:
Is this approved then @yihozhang ?

yihozhang (approved):

yihozhang:
Yes it looks good to me.

oflatt (approved):
Nice work!

PR #279, opened 2024-01-24
Interpret tests

### LINKED ISSUE #276: [Tree Optimizer] Integrate interpreter with tests
Make the tests call the interpreter to both test the interpreter and confirm that the "build" and "check" expressions are actually equal.

Alex-Fischman:
Changes `run_test` to interpret any expressions that are extracted from a test.

Also fixes the interpreter to fail to deal with functions correctly.

Closes #276


oflatt:
Could you add closes #276 to the PR description?

oflatt (changes requested):
We discussed ways to improve this PR. I think the current plan is to add another field to the test method, and we can add expressions to interpret there.

Alex-Fischman:
Closing because I will reimplement this on the new schema and starting from scratch is easier.

PR #280, opened 2024-01-24
Tree unique ast builder for rust

oflatt:
This PR adds a convenient API for constructing tree encoded terms in rust. It removes `Box` boilerplate and ids.
This will help with translation and testing in the future.


tree_unique_args/src/lib.rs
```diff
@@ -35,6 +36,11 @@ pub enum Expr {
     Or(Box<Expr>, Box<Expr>),
     Not(Box<Expr>),
     Get(Box<Expr>, usize),
+    /// Concat is a convenient built-in way
+    /// to put two tuples together.
+    /// It's not strictly necessary, but
+    /// doing it by constructing a new big tuple is tedius and slow.
+    Concat(Box<Expr>, Box<Expr>),
```
Alex-Fischman, line 43: Why is this an `Expr` instead of a `fn(Expr, Expr) -> Expr`?

tree_unique_args/src/lib.rs
```diff
@@ -44,9 +50,57 @@ pub enum Expr {
     Let(Id, Box<Expr>, Box<Expr>),
     Arg(Id),
     Function(Id, Box<Expr>),
+    /// A list of functions, with the first
+    /// being the main function.
+    Program(Vec<Expr>),
```
Alex-Fischman, line 55: A `Program` should not be an `Expr` variant.

Alex-Fischman (changes requested):

tree_unique_args/src/lib.rs
oflatt: This PR proposes adding a concat operator to our encoding.
It can't be a function on two exprs, since the two exprs haven't been evaluated yet. For example: `(Concat (Arg 0) (Arg 0))`

tree_unique_args/src/lib.rs
oflatt: I've made `Program` an `Expr` for the same reason `Function` is an `Expr`.
In this encoding, we've made everything an `Expr` for convenience.

Alex-Fischman: Function is not an Expr, that was a mistake that I made.

Alex-Fischman (approved):

tree_unique_args/src/schema.egg
```diff
@@ -68,7 +68,7 @@
 ; ==========================
 
 ;                   f      output
-(relation Function (IdSort Expr))
+(function Function (IdSort Expr) Expr)
```
ajpal, line 71: 👍

tree_unique_args/src/lib.rs
```diff
@@ -44,9 +50,57 @@ pub enum Expr {
...
     Call(Id, Box<Expr>),
 }
 
+impl Expr {
+    pub fn for_each_child(&mut self, mut func: impl FnMut(&mut Expr)) {
```
ajpal, line 60: What is `func`?
Could use a doc comment for this function

tree_unique_args/src/ast.rs
```diff
@@ -0,0 +1,204 @@
...
+            let new_id = *fresh_id;
+            *fresh_id += 1;
+            *id = Id(new_id);
+            give_fresh_ids_helper(body, new_id, fresh_id);
+        }
+        Call(id, arg) => {
+            *id = Id(current_id);
+            give_fresh_ids_helper(arg, current_id, fresh_id);
+        }
+        _ => expr.for_each_child(move |child| give_fresh_ids_helper(child, current_id, fresh_id)),
```
ajpal, line 37: What does `move` do?

tree_unique_args/src/interpreter.rs
```diff
@@ -50,6 +51,19 @@ pub fn typecheck(e: &Expr, arg_ty: &Option<Type>) -> Result<Type, TypeError> {
                 )),
             }
         }
+        Expr::Concat(tuple_1, tuple_2) => {
+            let ty_tuple_1 = typecheck(tuple_1, arg_ty)?;
+            let ty_tuple_2 = typecheck(tuple_2, arg_ty)?;
+            match (ty_tuple_1.clone(), ty_tuple_2) {
```
ajpal, line 57: Why do you only need a clone on the first?

ajpal (approved):
Couple questions, but lgtm!

tree_unique_args/src/ast.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+pub fn num(n: i64) -> Expr {
+    Num(n)
+}
+
+pub fn ttrue() -> Expr {
+    Boolean(true)
+}
+pub fn tfalse() -> Expr {
```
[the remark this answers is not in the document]
oflatt: I can't use `false` since it's a reserved keyword. I chose `t` for tree.
Is it an okay name? Why `ffalse`?

tree_unique_args/src/ast.rs
oflatt: `move` is a rust keyword that says "give this closure ownership of captured values"
I don't understand it completely but it makes this code type check

tree_unique_args/src/ast.rs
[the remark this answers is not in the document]
ajpal: Ah, `t` = tree. Fair enough. I just saw `ttrue` and didn't make the connection to trees.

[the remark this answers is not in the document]
Alex-Fischman: Rust also has the `r#` prefix for using reserved keywords, although it's arguable uglier than `t`.

tree_unique_args/src/ast.rs
Alex-Fischman: `move` gives the lambda ownership of the variables that it closes over. Here I would assume that it's necessary because `fresh_id` is a mutable reference.

tree_unique_args/src/lib.rs
oflatt: This method runs `func` on each child of `self`
Added a comment

tree_unique_args/src/interpreter.rs
oflatt: Good catch- we need to clone because we need the value later for the error message.
changed the code to clone both and more correct error messages

ajpal (approved):

PR #285, opened 2024-01-25
Add loops to type analysis

ajpal:


tree_unique_args/src/type_analysis.egg
```diff
@@ -271,4 +271,29 @@
...
+       (HasTypeDemand ctx (Loop id in pred-out))
+       (HasType ctx in ty)
+      )
+      (
+       (HasType id (Arg id) ty) ; assert the argument has type ty in the loop's context
+       (HasTypeDemand id pred-out) ; demand the types of the predicate and output in the loop's context
+      )
+      :ruleset type-analysis)
+
+(rule (
```
oflatt, line 291: This rule is good, but it doesn't give an error when the `pred-out` doesn't match the tuple format.
So I guess we need another rule?

oflatt (changes requested):

tree_unique_args/src/type_analysis.egg
ajpal: That'll probably get covered later when I add the rule that everything must get a type?
Or I can add an error rule here.

oflatt: It's true that you'll get an error that it didn't get a type...
I guess it'll be a pretty unhelpful error though. "Didn't get a type" should mean there's a bug in the type checker, while other errors mean the code doesn't typecheck

oflatt (approved):
Nice!

PR #287, opened 2024-01-25
fix function purity analysis

clyben:
Fix a small bug of purity of Call in purity analysis. Previously the purity of Call don't consider the purity of Function itself.


oflatt (approved):
Nice!

PR #288, opened 2024-01-25
Translate a single region from RVSDG to tree encoding

oflatt:
This PR implements the translation of a function with a single region to the tree encoding. It doesn't run the interpreter or optimizations on the result yet.

### DIFF
--- a/src/rvsdg/tree_unique/to_tree.rs
+++ b/src/rvsdg/tree_unique/to_tree.rs
@@ -38,11 +38,13 @@ struct RegionTranslator<'a> {
     /// to the index in the argument
     /// where the value is stored.
     /// After evaluating a node, do not evaluate it again.
-    /// Instead find it's index here.
+    /// Instead find its index here.
     index_of: HashMap<Id, usize>,
     nodes: &'a Vec<RvsdgBody>,
 }
 
+/// helper that binds a new expression, adding it
+/// to the environment using concat
 fn cbind(expr: Expr, body: Expr) -> Expr {
     tlet(concat(arg(), expr), body)
 }
@@ -53,7 +55,11 @@ impl<'a> RegionTranslator<'a> {
     fn add_binding(&mut self, expr: Expr, id: Id) -> usize {
         self.bindings.push(expr);
         let res = self.bindings.len() - 1 + self.num_args;
-        assert_eq!(self.index_of.insert(id, res), None);
+        assert_eq!(
+            self.index_of.insert(id, res),
+            None,
+            "Node already evaluated. Cycle in the RVSDG or similar bug."
+        );
         res
     }
 
--- a/src/rvsdg/tree_unique/to_tree.rs
+++ b/src/rvsdg/tree_unique/to_tree.rs
@@ -48,6 +48,7 @@ impl RvsdgFunction {
 
 #[test]
 fn simple_translation() {
+    use Expr::*;
     const PROGRAM: &str = r#"
     @sub() {
         v0: int = const 1;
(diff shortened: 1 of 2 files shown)


src/rvsdg/tree_unique/to_tree.rs
```diff
@@ -38,9 +38,11 @@ struct RegionTranslator<'a> {
     num_args: usize,
     /// a stack of let bindings to generate
     bindings: Vec<Expr>,
-    /// a map from the rvsdg node id
+    /// A map from the rvsdg node id
     /// to the index in the argument
-    /// where the value is stored
+    /// where the value is stored.
+    /// After evaluating a node, do not evaluate it again.
+    /// Instead find it's index here.
```
ajpal, line 41: nit: its

src/rvsdg/tree_unique/to_tree.rs
```diff
@@ -7,3 +7,245 @@
...
+    /// A map from the rvsdg node id
+    /// to the index in the argument
+    /// where the value is stored.
+    /// After evaluating a node, do not evaluate it again.
+    /// Instead find it's index here.
+    index_of: HashMap<Id, usize>,
+    nodes: &'a Vec<RvsdgBody>,
+}
+
+fn cbind(expr: Expr, body: Expr) -> Expr {
```
ajpal, line 46: Comment?
Makes a new let binding, concat is just a convenience to avoid lots of nesting, but doesn't change semantics (right?)

src/rvsdg/tree_unique/to_tree.rs
```diff
@@ -7,3 +7,245 @@
...
+    tlet(concat(arg(), expr), body)
+}
+
+impl<'a> RegionTranslator<'a> {
+    /// Adds a binding and returns its index
+    /// into the argument list.
+    fn add_binding(&mut self, expr: Expr, id: Id) -> usize {
+        self.bindings.push(expr);
+        let res = self.bindings.len() - 1 + self.num_args;
+        assert_eq!(self.index_of.insert(id, res), None);
```
ajpal, line 56: What is this assertion?

ajpal: (Could use a comment explaining what the invariant is / why it must hold / why it matters

tree_unique_args/src/ast.rs
```diff
@@ -1,5 +1,26 @@
 use crate::{Expr, Expr::*, Id, Order};
 
+impl Expr {
+    pub fn eq_ignoring_ids(&self, other: &Expr) -> bool {
+        let mut copy = other.clone();
+        give_fresh_ids(&mut copy);
+        self == &copy
```
ajpal, line 7: dumb q; why does this work?

src/rvsdg/tree_unique/to_tree.rs
oflatt: Yep, added a comment

src/rvsdg/tree_unique/to_tree.rs
oflatt: Added an error message that describes why it shouldn't occur

tree_unique_args/src/ast.rs
oflatt: Not a dumb question- I added a comment
```
    /// Check that two expressions are the same ignoring their ids.
    /// To do this, simply assign them all new ids.
    /// If they are the same expression, they will get the same ids
    /// since `give_fresh_ids` is deterministic.
```

ajpal (approved):

PR #289, opened 2024-01-25
No need for HasTypeDemand

ajpal:


oflatt (approved):
Looks good- we can resurrect if we decide we want to demand only some terms get types

PR #291, opened 2024-01-26
Replace unit with empty tuple and add id to All

rtjoa:
An `(All order (Nil))` doesn't have a referencing id, so we add one. We also remove unit as it's no longer needed.


oflatt (approved):
Wonderful!

PR #293, opened 2024-01-27
add ExprUsesArgs analysis

yihozhang:
This analysis returns a set of variables used by an expression. Users of this analysis need to explicitly place demands on the expression they want to run this analysis on.

### DIFF
--- a/tree_unique_args/src/arg_used_analysis.rs
+++ b/tree_unique_args/src/arg_used_analysis.rs
@@ -97,7 +97,7 @@ pub(crate) fn arg_used_analysis_rules() -> Vec<String> {
     ESort::iter()
         .map(|sort| {
             "
-            (function *UsesArgs (*) I64Set :merge (set-union old new))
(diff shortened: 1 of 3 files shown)


oflatt:
Why does this PR require using a set? Not sure what it is used for yet.
Could you do a relation `(relation ExprUsesVar (Expr Expr))`?

yihozhang:
There are several reasons for this:
* I need to collect a list `FV` of all free variables of an expression, and add `FV` to the input context of `let-inlining`. Collecting all values of a column in a container is less straightforward than just having a `set` as output.
* There was a typo in the original analysis. The merge function for `ExprUsesArgs` should be `set-intersect`, otherwise a constant may use variables if it is merged with `x-x`, etc. `relation` cannot do this.

oflatt:
So this is more like a `ExprAlwaysUsesArgs` analysis? As in, it uses these arguments in all equivalent versions?

oflatt:
I think we decided to close this, since we are changing it to extract a particular term that uses a particular set of arguments.

PR #295, opened 2024-01-28
Translate RVSDG loops to tree encoding 

oflatt:
Extends the translation from RVSDG to translate loops, and adds two tests.
Makes some convenient macros for constructing asts.

### DIFF
--- a/src/rvsdg/tree_unique/to_tree.rs
+++ b/src/rvsdg/tree_unique/to_tree.rs
@@ -61,6 +61,8 @@ impl<'a> RegionTranslator<'a> {
         res
     }
 
+    /// Make a new translator for a region with
+    /// num_args and the given nodes.
     fn new(num_args: usize, nodes: &'a Vec<RvsdgBody>) -> RegionTranslator {
         RegionTranslator {
             num_args,
@@ -70,10 +72,8 @@ impl<'a> RegionTranslator<'a> {
         }
     }
 
-    /// Build a translator and translate
-    /// the operands to the tree encoding.
-    /// Produces a tree-encoded term that evaluates
-    /// to a tuple containing results.
+    /// Wrap the given expression in all the
+    /// bindings that have been generated.
     fn build_translation(&self, inner: Expr) -> Expr {
         let mut expr = inner;
 
--- a/src/rvsdg/tree_unique/to_tree.rs
+++ b/src/rvsdg/tree_unique/to_tree.rs
@@ -9,18 +9,22 @@
 //! program.
 #[cfg(test)]
 use crate::{cfg::program_to_cfg, rvsdg::cfg_to_rvsdg, util::parse_from_string};
+#[cfg(test)]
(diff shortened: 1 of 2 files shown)


src/rvsdg/tree_unique/to_tree.rs
```diff
@@ -55,37 +61,39 @@ impl<'a> RegionTranslator<'a> {
...
-        for binding in translator.bindings.into_iter().rev() {
-            expr = cbind(binding, expr);
+        for binding in self.bindings.iter().rev() {
+            expr = cbind(binding.clone(), expr);
         }
         expr
     }
 
-    fn translate_operand(&mut self, operand: Operand) -> usize {
+    /// Returns a pure expression (e.g. `getarg(0)`) that
```
ajpal, line 86: I don't understand the e.g. What is getarg?

src/rvsdg/tree_unique/to_tree.rs
```diff
@@ -55,37 +61,39 @@ impl<'a> RegionTranslator<'a> {
...
+    fn translate_operand(&mut self, operand: Operand) -> Expr {
         match operand {
-            Operand::Arg(index) => index,
-            Operand::Id(id) => self.translate_node(id),
-            Operand::Project(_id, _indexx) => {
-                todo!("Doesn't handle subregions yet");
+            Operand::Arg(index) => getarg(index),
+            Operand::Id(id) => getarg(self.translate_node(id)),
+            Operand::Project(p_index, id) => {
+                let index = self.translate_node(id);
```
ajpal, line 95: Why is this called index? Isn't this a region?

src/rvsdg/tree_unique/to_tree.rs
```diff
@@ -135,8 +166,9 @@ impl<'a> RegionTranslator<'a> {
             BasicExpr::Print(args) => {
                 assert!(args.len() == 2, "print should have 2 arguments");
                 let arg1 = self.translate_operand(args[0]);
+                // should be unit
```
ajpal, line 169: What should be unit?

ajpal (approved):
couple questions, but lgtm!

src/rvsdg/tree_unique/to_tree.rs
oflatt: `getarg` is a helper that desugars to `(Get (Arg (Id 0)) 0)`

src/rvsdg/tree_unique/to_tree.rs
oflatt: The whole region becomes a let-bound tuple, so this index is the index of that tuple in the environment 
Added some comments that explain this better

src/rvsdg/tree_unique/to_tree.rs
oflatt: Added a comment- argument 2 is the incoming print edge state, which we translate to unit in the tree encoding.

PR #296, opened 2024-01-29
Tree schema that supports shared and unique

oflatt:
This PR adds a tree schema that can support both shared terms and unique terms.
Compared to `tree_unique`:
- Adds a new variant of `Id` called `Shared`, which can be used for shared terms
- Adds a test in the id analysis to prevent mixing shared and unique ids in terms
- Adds ids to branches of switch statements, allowing for path-condition based optimizations. `Switch` statements must now have `Branches` as children. Adds a test that ensures this.
- Deletes most optimizations, but keeps most utility
- Adds `IsValidShared` and `IsValidUnique`, subsets of `IsValid` for shared and not-shared optimizations.
`IsValid = IsValidShared \union IsValidUnique`


rtjoa (approved):
Looks good (just skimmed).

Note: Before people start hacking on their optimizations, I think we should do follow-ups on this that implement typed args and lets/loops-are-lambdas.

Nit: I think the directory should be called `tree` rather than `tree_shared`. Then, once all optimizations are ported, we'll delete `tree_unique_args`.

PR #297, opened 2024-01-29
Add Read/Write to Type analysis

ajpal:
Adds type annotation to Read.


tree_unique_args/src/type_analysis.rs
```diff
@@ -1,3 +1,6 @@
+#[cfg(test)]
+const SCHED: &str = "(run-schedule (repeat 5 (saturate type-analysis) (saturate always-run)))";
```
oflatt, line 2: Do you need this schedule?
I think we should add all your typechecking rules to the normal schedule, and `run_test` will just work

oflatt (changes requested):
Great! Had a question about schedules

oflatt (approved):
Missed your comment, nice work

PR #298, opened 2024-01-29
Add typing rules for functions/calls

ajpal:


tree_unique_args/src/type_analysis.egg
```diff
@@ -211,4 +211,25 @@
...
       )
       ((panic "Loop did not get two arguments (predicate and output)"))
+      :ruleset type-analysis)
+
+; Functions
+(rule ((= f (Function id body in-ty out-ty)))
+      ((HasType (Arg id) in-ty))
+      :ruleset type-analysis)
+
+(rule (
```
oflatt, line 221: We should probably have a rule that errors when out-ty for the function and out-ty for the body are different

oflatt (approved):
Great!

PR #302, opened 2024-01-30
fix switch bug

ajpal:
`if x then y else z` is the same as `switch x (cons z (cons y nil))` (note the switched branches)
We've made this mistake a few times and I think fixed it on various branches, but we should just merge the change to main.


Alex-Fischman (approved):
LGTM

Alex-Fischman:
Actually the 3/4 optimizations are also wrong

PR #303, opened 2024-01-31
Add If to schema

oflatt:
This PR adds `If` so that we can differentiate it from `Switch`. Now when the predicate is a boolean, `If` must be used.


ajpal (approved):
thanks for adding! lgtm

PR #305, opened 2024-01-31
Initial wiring for memory into RVSDGs

ezrosent:
The structure for these memory operations is extremely similar to prints, with all operations "listening" to the same state edge introduced for prints.

This PR does not touch any of the conversions to and from egglog in either encoding.


src/rvsdg/from_cfg.rs
```diff
@@ -421,11 +421,20 @@ impl<'a> RvsdgBuilder<'a> {
...
+                    ValueOps::Load | ValueOps::Alloc => {
+                        // NB: Load takes a state edge _and_ returns one. We
+                        // could relax this later (making it easier to reorder
+                        // loads, etc.), but only after confirming that it is
+                        // safe wrt the model we're using. For example, in
+                        // WebAssembly, loads can trap.
+                        let dest_var = self.analysis.intern.intern(dest);
+                        let mut ops = convert_args(args, &mut self.analysis, &mut self.store, pos)?;
+                        ops.push(self.store[&self.analysis.state_var]);
+                        let expr = BasicExpr::Op(*op, ops, op_type.clone());
```
oflatt, line 433: Should `Load` be another variant of `BasicExpr`? It might require special handling due to state edges unlike other ops

oflatt (changes requested):
Nice work, I had a small question.
I think at this point we can test `RvsdgConversion` on bril tests with memory?
Could be worth adding that to the testing infra (maybe by adding a `supports_memory` predicate on the run modes)

src/rvsdg/from_cfg.rs
ezrosent: What kind of special handling does would that be? I think most kinds of special handling could happen by pattern matching, what does the extra variant buy?

ezrosent:
Sorry, I'm not quite sure what you're asking here. How would we compute the value of the "supports_memory" filter? AIUI the desired flow would be:
* Add new bril / insta tests for memory.
* Run all the memory-involving tests on modes where supports_memory=true (just RvsdgConversion for now?)

Would it be easier to add those tests once memory operations are wired into bril, etc? (I'm also not against GCing more of the old run modes too). Then we could just run them uniformly everywhere, with unit tests ensuring we have test coverage until then.

oflatt:
Yeah, I'm proposing adding in some tests for memory from Bril's test suite and just run `RvsdgConversion` on them. In the past, running on these files in addition to unit tests has uncovered lots of bugs.

I agree it would be easier if all run modes supported memory, but it'll be a while until round-trip with memory works.
Yeah, I'm hoping (real soon!) we can delete all code not related to our favorite schema.

ezrosent:
Added the memory tests from bril. This did uncover a bug in SVG generation that I patched!

src/rvsdg/from_cfg.rs
oflatt: you are right

oflatt (approved):
great work!

PR #306, opened 2024-01-31
Add an `If` variant to Rvsdg struct

oflatt:
Currently `Switch` works on both booleans and integers.
This PR changes the Rvsdg struct to have an `If` that works on booleans and a `Switch` that works on integers.
This change required a significant amount of refactoring. For the translation to and from the milestone encoding, the current behavior is preserved by using `Switch`.


src/rvsdg/mod.rs
```diff
@@ -129,9 +129,17 @@ pub(crate) enum Operand {
 pub(crate) enum RvsdgBody {
     BasicOp(BasicExpr<Operand>),
 
+    /// Conditional branch, witha boolean predicate.
```
ezrosent, line 132: with a

ezrosent (approved):

PR #315, opened 2024-02-02
Assume Schema- New Schema Solves Context with Assume

oflatt:
Using assume nodes to create context is much cleaner
Also, tuple values are flat


tree_inputs/src/schema.egg
```diff
@@ -0,0 +1,196 @@
...
+; =================================
+
+(datatype Assumption
+  ; The argument is bound to `expr`.
+  ; The argument cannot be unioned with `expr` because `expr`
+  ; is evaluated once at the nearest enclosing `Let`.
+  ;      expr
+  (InLet Expr)
+  ; The term is in a loop with `input` and `output`.
+  ;      input    output
```
clyben, line 169: should it be `output` or `pred_out`?

oflatt: fixed

ezrosent (approved):
(did a review over zoom and discussed this one. LGTM!)

PR #316, opened 2024-02-02
Assume schema rust definition

oflatt:


tree_inputs/src/schema.rs
```diff
@@ -0,0 +1,71 @@
...
+//! be implemented in this file.
+
+use std::rc::Rc;
+
+pub enum Type {
+    IntT,
+    BoolT,
+    UnitT,
+    /// Nested tuple types are not allowed.
+    TupleT(Vec<Rc<Type>>),
```
ezrosent, line 12: If nested tuples aren't allowed, can we make them unrepresentable, as in:

```rust
pub enum BaseType { Int, Bool, Unit }
pub enum Type { Prim(BaseType), Tuple(Vec<BaseType>) }
```

This way we also get to avoid boxing the tuple elements.

oflatt: Good idea

oflatt:
closing for now, will open a new PR for this

PR #317, opened 2024-02-02
Input schema rust interpreter

oflatt:


oflatt:
Trying again now that we changed to asserts.

PR #323, opened 2024-02-06
Assume schema in rust

oflatt:
See diff here:
https://github.com/egraphs-good/eggcc/compare/main..oflatt-assume-rust
Github uses "three dot" diff by default but we need a "two dot" diff to just see changes

### DIFF
--- a/tree_assume/src/schema.rs
+++ b/tree_assume/src/schema.rs
@@ -1,0 +1,87 @@
+//! This module mirrors `schema.egg`.
+//! No implementation or conversion should
+//! be implemented in this file.
+
+use std::rc::Rc;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BaseType {
+    IntT,
+    BoolT,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Type {
+    Base(BaseType),
+    PointerT(BaseType),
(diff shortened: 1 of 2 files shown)


tree_assume/src/schema.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Order {
+    Parallel,
+    Sequential,
+    Reversed,
+}
+
+pub type RcExpr = Rc<Expr>;
```
oflatt, line 53: add comment for why Rc

tree_assume/src/schema.rs
```diff
@@ -0,0 +1,87 @@
+//! This module mirrors `schema.egg`.
+//! No implementation or conversion should
+//! be implemented in this file.
+
+use std::rc::Rc;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BaseType {
+    IntT,
+    BoolT,
```
oflatt, line 10: add comment for why no unit type

tree_assume/src/schema.rs
ezrosent, line 53: Maybe explain "Why Rc" here.

ezrosent (approved):
Change looks good, modulo fixing the git history on this PR.

PR #326, opened 2024-02-06
Assume schema conversion to egglog

oflatt:
Diff:
https://github.com/egraphs-good/eggcc/compare/oflatt-assume-interpreter..oflatt-assume-run-test


oflatt:
close in favor of clean git hist

PR #329, opened 2024-02-06
Convert tree exprs to egglog encoding

oflatt:


tree_assume/src/schema.egg
```diff
@@ -112,7 +112,7 @@
 
 ; global short hands for orders
 (let par (Parallel))
-(let seq (Sequential))
+(let sequ (Sequential))
```
ajpal, line 115: why?

ajpal (approved):

PR #330, opened 2024-02-07
Start Type analysis

ajpal:


tree_assume/src/type_analysis.egg
```diff
@@ -0,0 +1,100 @@
...
+(rule ((= lhs (Const (Bool b))))
+      ((HasType lhs (Base (BoolT))))
+      :ruleset type-analysis)
+
+(rule ((= lhs (Empty)))
+      ((HasType lhs (TupleT (TNil))))
+      :ruleset type-analysis)
+
+; Unary Ops
+(rule (
```
oflatt, line 19: I have an idea for how to refactor this rule to cover more erroring cases. How about something like:
```
(rule ((= lhs (Uop (Not) e)))
        ((HasType e (Base (BoolT)))
         (HasType lhs (Base (BoolT)))))
```
This enforces both that the input has type bool and that the output does.
So you'll get a type error for this program: `(Uop (Not) (Int 2))`
Which you won't get an error for right now I think

tree_assume/src/type_analysis.egg
```diff
@@ -0,0 +1,100 @@
...
+      ((HasType lhs (Base (BoolT))))
+      :ruleset type-analysis)
+
+(rule ((= lhs (Uop (Print) e)))
+      ((HasType lhs (TupleT (TNil))))
+      :ruleset type-analysis)
+
+(rule (
+        (= lhs (Uop (Load) e))
+        (HasType e (PointerT ty))
```
oflatt, line 32: This one you can't play the same trick, so I guess you'll have to write something like `(relation ExpectPointerType (Type))` and write some rules that make sure no other types get in there

tree_assume/src/type_analysis.egg
```diff
@@ -0,0 +1,100 @@
...
+        (HasType e1 (Base (BoolT)))
+        (HasType e2 (Base (BoolT)))
+      )
+      ((HasType lhs (Base (BoolT))))
+      :ruleset type-analysis)
+
+(rule (
+        (= lhs (Bop (Write) ptr val))
+        (HasType ptr (PointerT ty))
+        (HasType val (Base t)) ; must be BaseType??
```
oflatt, line 89: I'm not sure, can you write pointers to memory? Certainly not tuples.

tree_assume/src/type_analysis.egg
```diff
@@ -0,0 +1,165 @@
...
+        (HasType e (Base (BoolT)))
+      )
+      ((HasType lhs (Base (BoolT))))
+      :ruleset type-analysis)
+(rule ((= lhs (Uop (Not) e)))
+      ((ExpectType e (Base (BoolT)) "(Not) expected bool, received {?}"))
+      :ruleset type-analysis)
+
+
+(rule ((= lhs (Uop (Print) e)))
```
oflatt, line 41: We should ensure that the child has some type, even though we don't care what type it is.
Would be great if we checked that it wasn't a tuple type but that's hard. add a TODO for non-tuple checking

oflatt (approved):
Awesome work

PR #332, opened 2024-02-07
Simple constant folding optimization as an example

oflatt:


oflatt:
I considered separating these two, but we should almost always be calling both. This way, you won't forget to.

rtjoa:
> I considered separating these two, but we should almost always be calling both. This way, you won't forget to.

I think tests will come out cleaner if they are separate. If it's good style to test both, this should happen during review.

Also note that it's odd we expect all test programs to have the same expected value - turning the param from a vector of programs to a single program (or: vector of program-output pairs) will encourage more diverse testing.

oflatt:
We should add that too, but a lot of our tests involve strait up egglog code (f.e. analysis tests)
Having a list of values would also be nice but for now this is good enough

ajpal:
> > I considered separating these two, but we should almost always be calling both. This way, you won't forget to.
> 
> I think tests will come out cleaner and be more readable if they are separate. If it's good style to test both, this should happen during review.
> 
> Also note that it's odd we expect all test programs to have the same expected value - turning the param from a vector of programs to a single program (or: vector of program-output pairs) will encourage more diverse testing.

+1

oflatt:
Chatted with @ajpal about this
For now lets go with the existing one, we can add @rtjoa's `egglog_unit_test_opt` as needed

PR #333, opened 2024-02-08
Add rules that propogate assume nodes to create unique contexts

oflatt:
This PR adds some elegant rules that propagate assume nodes down from a top-level function. It also adds a couple of tests.

Not included is rules for `Switch`, which are a little harder.

This PR changes `Assumption` to have *three* new variants: `InIf` and `InFunction`, and `InSwitch`.
I realized that our previous plan of using a strait up boolean condition isn't quite right due to effects. For example, if an `If` statement reads from memory, the value it reads is whatever was in memory at the execution time of the predicate.

### DIFF
--- a/tree_assume/src/utility/assume.egg
+++ b/tree_assume/src/utility/assume.egg
@@ -1,0 +1,111 @@
+; This file propogates assume nodes top-down from functions.
+; It gives each program path a unique equality relation.
+; This can be quite expensive, so be careful running these rules.
+
+(ruleset assume)
+
+(sort AssumeList)
+; We collect multiple assumes into one AssumeMulti.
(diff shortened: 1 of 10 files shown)


tree_assume/src/utility/assume.egg
```diff
@@ -0,0 +1,81 @@
...
+         :ruleset assume)
+
+; ################### start top-down assumptions
+
+(rewrite
+ (Function name in_ty out_ty out)
+ (Function name in_ty out_ty
+   (Assume (InFunc  name)
+               out))
+ :ruleset assume)
```
rtjoa, line 24: Per discussion on Slack, I would just remove this rule for now

rtjoa (approved):

PR #336, opened 2024-02-09
[Tree assume] Add metaprogramming tooling

rtjoa:
Add metaprogramming similar tree-unique's `ir.rs` to tree-assume.

---

PR stack:
1. #336
2. #337 
3. #338 
3. #339


oflatt (approved):
Looks good to me, don't forget to remove dead code annotation when it's in use

PR #337, opened 2024-02-09
[Tree assume] Print schema on test failure and `cargo run`

rtjoa:
Improve test debugging workflow by printing both schema and error on test failure, and printing schema on `cargo run`.

---

PR stack:
1. #336
2. #337 
3. #338 
3. #339

PR #338, opened 2024-02-09
[Tree assume] Add `IsValid`

rtjoa:
Add `IsValid` to tree-assume. See #262.

---

PR stack:
1. #336
2. #337 
3. #340 
4. #338 
3. #339

### DIFF
--- a/tree_assume/src/optimizations/is_valid.rs
+++ b/tree_assume/src/optimizations/is_valid.rs
@@ -1,0 +1,65 @@
+use crate::schema_helpers::{Constructor, ESort, Purpose};
+use strum::IntoEnumIterator;
+
+fn rule_for_ctor(ctor: Constructor) -> Option<String> {
+    let actions = ctor.filter_map_fields(|field| match field.purpose {
+        Purpose::Static(_) => None,
+        Purpose::CapturedExpr | Purpose::SubExpr | Purpose::SubListExpr => Some(format!(
+            "({sort}IsValid {var})",
+            sort = field.sort().name(),
+            var = field.var()
+        )),
+    });
+
+    if actions.is_empty() {
+        return None;
+    }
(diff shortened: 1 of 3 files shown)


tree_assume/src/optimizations/is_valid.rs
```diff
@@ -0,0 +1,65 @@
...
+(fail (check (ExprIsValid {not_made_valid})))
+    ",
+        num0 = int(0),
+        arg = arg(),
+        tup45 = parallel!(int(4), int(5)),
+    );
+    crate::egglog_test(
+        &build,
+        &check,
+        vec![],
```
oflatt, line 61: Pass the loop in here to check that it runs

tree_assume/src/optimizations/is_valid.rs
```diff
@@ -0,0 +1,65 @@
...
+    ESort::iter()
+        .map(|sort| "(relation *IsValid (*))".replace('*', sort.name()))
+        .chain(Constructor::iter().filter_map(rule_for_ctor))
+        .collect::<Vec<_>>()
+}
+#[cfg(test)]
+use crate::ast::*;
+#[cfg(test)]
+use crate::Value;
+
```
oflatt, line 35: Might be good to have a test that shows that validity passes through the assume node but not its left child

oflatt (changes requested):
Looks good, had a couple suggestions

tree_assume/src/optimizations/is_valid.rs
rtjoa: Updated

tree_assume/src/optimizations/is_valid.rs
rtjoa: Done

oflatt (approved):
great!

PR #340, opened 2024-02-09
[Tree assume] Fix get's to_egglog

rtjoa:
Get doesn't take a const, it just takes a literal i64.

This behavior is exercised in a follow-up PR.

---

PR stack:
1. #336
2. #337 
3. #340 
4. #338 
3. #339


oflatt (approved):
good catch

PR #341, opened 2024-02-09
Add `Subst` rule that creates a new context as it substitutes.

oflatt:


kirstenmg (approved):
Looks good as far as I can tell!

PR #346, opened 2024-02-12
Update rust and egglog versions in tree assume

kirstenmg:
A dependency of tree assume requires at least 1.74. egglog needed to be udpated as well.


oflatt (approved):
Thanks for fixing!

PR #348, opened 2024-02-12
[Tree Assume] Add type to Arg

ajpal:


oflatt (approved):
Looks great
We can consider sugar for arg(intt()) in the future

PR #349, opened 2024-02-12
[Tree Assume][Type Analysis] Let

ajpal:


oflatt (approved):
nice work
for posterity, we decided on top-down  rules so that we don't have to worry about the case where the body does not reference arg

PR #353, opened 2024-02-13
Memory ops for RVSDG => Bril

### LINKED ISSUE #343: RVSDG with memory -> Bril translation

ezrosent:
Wire in memory ops in to_cfg. Opening a draft as there are a few questions in to_cfg I'd like to resolve.

Fixes #343

### DIFF
--- a/src/rvsdg/to_cfg.rs
+++ b/src/rvsdg/to_cfg.rs
@@ -798,11 +798,11 @@ impl<'a> RvsdgToCfg<'a> {
                 }
             }
             BasicExpr::Effect(op, args) => {
-                // TODO: should we combine these?
                 let mut results: Vec<TranslationResult> = args
                     .iter()
                     .map(|op| self.operand_to_bril(*op, current_args, ctx))
                     .collect();
+                // Combine all arguments but the state edge.
                 let main_result = self.combine_results(&results[0..results.len() - 1]);
                 let args: Vec<String> = main_result
                     .values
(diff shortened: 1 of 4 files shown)


src/rvsdg/to_cfg.rs
```diff
@@ -763,33 +797,35 @@ impl<'a> RvsdgToCfg<'a> {
...
-                assert!(print_operands.len() == 2);
-                let argument = self.operand_to_bril(print_operands[0], current_args, ctx);
-                // also need to evaluate other prints before this one
-                let second_evaluated = self.operand_to_bril(print_operands[1], current_args, ctx);
-
-                let instructions = vec![Instruction::Effect {
-                    op: EffectOps::Print,
-                    args: vec![argument.get_single_res().unwrap_name()],
+            BasicExpr::Effect(op, args) => {
+                // TODO: should we combine these?
```
oflatt, line 801: remove todo, comment that we combine all but the state edge

oflatt (changes requested):
Nice work!
Is this tested by the current infra? I think you need to say that the RVSDG roundtrip run mode supports memory.

ezrosent:
Done! This uncovered a preexisting issue with converting empty functions:

```
@funcA(p: ptr<int>) {
  ret;
}

@main {
  five: int = const 5;
  x: ptr<int> = alloc five;
  call @funcA x;
  free x;
}
```

Gets translated back correctly, but `@funcA` had an empty body

```
@funcA(p: ptr<int>) {
}
```

Which caused the bril interpreter to panic. I added some code to fix that.

oflatt (approved):
Great!

PR #355, opened 2024-02-15
Start loops

ajpal:
More to come including tests for error cases, but this was already kind of tricky so it's probably worth starting review now.

### DIFF
--- a/tree_assume/src/schedule.egg
+++ b/tree_assume/src/schedule.egg
@@ -4,5 +4,8 @@
     (repeat 100 assume)
     (saturate always-run)
     (saturate error-checking)
-    (saturate type-analysis)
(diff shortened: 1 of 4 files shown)


tree_assume/src/schedule.egg
```diff
@@ -4,5 +4,8 @@
     (repeat 100 assume)
     (saturate always-run)
     (saturate error-checking)
-    (saturate type-analysis)
+    (repeat 10 ; TODO: do this til saturation?
```
oflatt, line 7: We discussed that this can be a saturate
I would use a different name than "immediate"
Maybe type-helpers?

oflatt (approved):
Looks great! Had one small request, but you can merge after

PR #364, opened 2024-02-19
Don't generate let bindings for pure expressions

oflatt:
This PR modifies the translation from RVSDG to tree encoding to not generate let bindings for pure expressions.

### DIFF
--- a/src/rvsdg/tree_unique/to_tree.rs
+++ b/src/rvsdg/tree_unique/to_tree.rs
@@ -30,13 +30,16 @@ use tree_in_context::{
 };
 
 impl RvsdgProgram {
-    pub fn to_tree_encoding(&self) -> TreeProgram {
+    /// Converts an RVSDG program to the tree encoding.
+    /// When `optimize_lets` is true, the conversion will also
+    /// try to prevent adding unnecessary let bindings.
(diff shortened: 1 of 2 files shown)


src/rvsdg/tree_unique/to_tree.rs
```diff
@@ -103,6 +114,17 @@ fn bind_value(expr: RcExpr, body: RcExpr) -> RcExpr {
 }
 
 impl<'a> RegionTranslator<'a> {
+    /// Adds a pure expression to the cache.
+    fn add_pure_value(&mut self, expr: RcExpr, id: Id) -> StoredNode {
+        if self.optimize_lets {
+            let res = StoredValue::Expr(expr.clone());
```
ezrosent, line 120: It'd help to have a comment here explaining what's happening.

It sounds like we are essentially "inlining" the expression rather than binding it, is that right?

ezrosent (approved):
Change is pretty straightforward, nice!

src/rvsdg/tree_unique/to_tree.rs
oflatt: Yes, we are inlining instead of binding. I'll add a comment

PR #367, opened 2024-02-19
Add tuple-ith and tuple-length

clyben:
Use type information of tuple to get tuple actual length, then get the i-th of a Concat by i-th of sub-tuples.


oflatt (approved):
Looks good

PR #373, opened 2024-02-21
Add tuple-remove-fst, DoWhile-output , DoWhile-output-ith.

clyben:
tuple-remove-fst take a tuple and return a tuple that the first Non-Tuple(TNil) typed term is removed.
DoWhile-output give the output tuple of a loop, which is the body with pred removed.
DoWhile-output-ith give the ith-output of a loop.


tree_in_context/src/utility/util.egg
```diff
@@ -35,3 +35,37 @@
...
+       (= tup (Concat ord expr1 expr2))
+       (HasType expr1 (TupleT (TCons ty tl))))
+      ((union (tuple-remove-fst tup) (Concat ord (tuple-remove-fst expr1) expr2))) :ruleset always-run)
+
+(rule ((tuple-remove-fst tup)
+       (= tup (Concat ord expr1 expr2))
+       (HasType expr1 (TupleT (TNil))))
+      ((union (tuple-remove-fst tup) (Concat ord expr1 (tuple-remove-fst expr2)))) :ruleset always-run)
+
+(function DoWhile-outputs-ith (Expr i64) Expr :unextractable)
```
oflatt, line 59: Are these helpers necessary? They create extra rows in the database.
Maybe the rule writer should just use tuple-ith

tree_in_context/src/utility/util.egg
```diff
@@ -35,3 +35,37 @@
...
+(function DoWhile-outputs (Expr) Expr :unextractable)
```
oflatt, line 60: Same here- just use tuple-remove-fst

tree_in_context/src/utility/util.egg
```diff
@@ -35,3 +35,37 @@
...
+(function DoWhile-pred (Expr) Expr :unextractable)
```
oflatt, line 61: Use tuple-ith 0 instead

oflatt (approved):
Looks good to me, I think we don't need the extra helpers

tree_in_context/src/utility/util.egg
```diff
@@ -35,3 +35,37 @@
        (= len1 (tuple-length expr1))
        (= ith (tuple-ith expr2 i)))
        ((union (tuple-ith (Concat ord expr1 expr2) (+ len1 i)) ith)) :ruleset always-run)
+
+(function tuple-remove-fst (Expr) Expr :unextractable)
+(rule ((tuple-remove-fst tup)
+       (= tup (Concat ord (Single expr1) expr2)))
```
yihozhang, line 41: Is it possible that the first argument to `Concat` is `(Let ... (Single ...))`?

yihozhang: I have the same question about  `tuple-ith`. We probably had the assumption that  a tuple is always a tree of uninterrupted `Concat`s and `Single`s. Would be nice to enforce that in our type system, or leave a note somewhere.

clyben:
Actually I think I don't need tuple-remove-fst for now. Shall I close it?

oflatt:
sure

PR #376, opened 2024-02-22
[tree in context] Function inlining

kirstenmg:
Currently, function inlining passes the provided tests when in_context is not saturated. However, with recursive functions, function inlining blows up too much and causes the in_context to take forever to saturate, so the tests do not terminate even after a minute.

Marking as draft until the timeout is resolved - either by modifying how context propagates, or limiting inlining.

### DIFF
--- a/tree_in_context/src/typechecker.rs
+++ b/tree_in_context/src/typechecker.rs
@@ -242,7 +242,7 @@ impl<'a> TypeChecker<'a> {
                 let func = self
                     .program
                     .get_function(string)
-                    .expect(format!("Function {string} should exist in program.").as_str());
+                    .unwrap_or_else(|| panic!("Function {string} should exist in program."));
                 assert_eq!(
                     aty,
                     func.func_input_ty().unwrap(),


yihozhang (commented):
LGTM! I have a question though, with the new `LoopArg`/`LetArg` PR (#383), do you need to update your optimization to make sure `f-output` uses `LetArg` instead of `LoopArg`?

kirstenmg:
Yes, I will need to replace all `(FuncScope)` in `f-output` with `(LetScope)`. This means `f-output` may contain old `InContext`s with `(FuncScope)` args, so function inlining may rely on substitution now to replace these old contexts?

yihozhang:
Yeah, we need to do something like this.

oflatt:
I'll be working on a substitution that can replace arguments that refer to the function scope

oflatt:
I believe this is now un-blocked by #392

kirstenmg:
I needed to update the type analysis for it to work for the function inlining tests:
In `type_analysis.egg`, I added a rule for `GreaterThan` and removed a line from the `Call` type rule that prevented recursive functions from getting typed (added to `HasType`).
In `typechecker.rs`, I added a more informative error message for a call to a nonexistent function. This was useful for debugging.

I included these changes in this PR, rather than their own PR, since the function inlining tests serve as tests for these changes.

oflatt (approved):
Great work! Love the testing improvement

PR #379, opened 2024-02-23
Create new contexts with only context-free children

oflatt:
Creating new contexts is a subtle thing to get right.
This PR proposes that the automatic rules only create contexts with context-free children. Without this change, rules never saturate even for non-cyclic egraphs due to InLoop.

We should keep thinking about what contexts to create and how rules can indicate contexts to create.

### DIFF
--- a/tree_in_context/src/schedule.egg
+++ b/tree_in_context/src/schedule.egg
@@ -1,11 +1,14 @@
 (run-schedule
+  (saturate
+   (saturate always-run)
+   context_creation
+   (saturate context_propagate))
   (repeat 6
-    (saturate always-run)
-    (repeat 20 in_context)
     (saturate always-run)
     (saturate error-checking)
     (saturate
       (saturate type-helpers)
       type-analysis
     )
-    (saturate constant_fold)))
+    (saturate constant_fold)
+    loop-simplify))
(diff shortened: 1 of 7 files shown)


tree_in_context/src/utility/in_context.egg
```diff
@@ -2,54 +2,41 @@
...
 (sort InContextList)
 
-; In order to saturate and not create unecessary contexts, we need to collapse
-; duplicate assumptions.
-; For example, the rewrite over `Function` could create many
-; `(InContext (InFunc name) (InContext (InFunc name) ...))` nestings
-(rewrite (InContext assumption (InContext assumption rest))
+; Here's a helpful rule that chooses a more specific context over
+; an inner, less specific one.
+(rewrite (InContext assumption (InContext assumption2 rest))
```
clyben, line 12: Will this rule still fire if we ban creating new InContext when subfield have InContext?

clyben (approved):
Looks good! Have tiny questions.

tree_in_context/src/utility/in_context.egg
oflatt: It may still fire if other rules create contexts

tree_in_context/src/schedule.egg
```diff
@@ -1,11 +1,14 @@
 (run-schedule
+  (saturate
+   (saturate always-run)
```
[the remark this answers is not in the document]
oflatt: I'm leaving it for now so we can use sugar if we want to

PR #380, opened 2024-02-23
Three quarters switch rewrites

Alex-Fischman:
This PR adds switch rewrites and switch rewrite tests. It also fixes an interpreter bug with short circuiting.


oflatt (approved):
great work

PR #381, opened 2024-02-23
[Tree in context] Add CICM rules

rtjoa:
Performs the CICM rules version in previous encodings, which perform rewrites like the following (arg types elided for brevity):
```
(Switch (Const (Int 1)) '(
  (Bop (LessThan) (Get (Arg ...) 0) (Const (Int 7)))
  (Bop (LessThan) (Get (Arg ...) 1) (Const (Int 7)))))

~~>

(Bop (LessThan)
  (Switch (Const (Int 1)) '(
    (Get (Arg ...) 0)
    (Get (Arg ...) 1)))
  (Const (Int 7)))
```

We'll eventually want to handle more complicated examples that require us to "reach through" lets.


oflatt (approved):
Nice work- are you going to make this code work for `If` as well?

rtjoa:
> Nice work- are you going to make this code work for `If` as well?

Thanks for the review. Will follow up with this!

PR #389, opened 2024-02-29
[Type Analysis] Assert that programs get a type from the type analysis

ajpal:


tree_in_context/src/lib.rs
```diff
@@ -55,6 +55,46 @@ pub fn optimize(program: &TreeProgram) -> std::result::Result<TreeProgram, egglo
...
 }
 
+fn check_func_type(func: RcExpr) {
+    let prologue = [
+        include_str!("schema.egg"),
+        include_str!("type_analysis.egg"),
+    ]
+    .join("\n");
+    let schedule = "(run-schedule
+      (repeat 6
```
oflatt, line 65: Extra unnecessary repeat 6 here

oflatt (approved):
Awesome!

PR #390, opened 2024-03-01
Context-sensitive substitution with cycle detection 

oflatt:
This PR adds new substitution rules that perform a substitution, copying all equalities to a new context.
However, the main challenge is that cycles can occur when performing this copying. This PR solves the issue by detecting these cycles and *unioning* the two resulting contexts.

It can therefore copy over cycles to the new equivalent relation as well, and figure out when two contexts are equivalent due to cycles in the original egraph.


oflatt:
I realized that the rules in this PR are not *quite* right
I have an idea to fix it, marking as a draft for now.

oflatt:
The version I just pushed seems good
There may be some saturation issues lingering when you try to substitute the same thing twice

oflatt:
Marked as ready- added more tests of cyclic graphs

rtjoa (approved):
Good ideas. Approving in case this PR is needed to unblock anyone. That said:
- It would be preferable to merge after removing loop args since there are currently some subtle choices in rules to do with different arg types (that I still don't completely understand).
- From what I can tell, substitution and context creation are combined because they both use cycle detection. Is this true? If so, I think could be a better programming interface, and give us more control (e.g. allow us to do "surgical" context insertion), if they were separated and we factored out cycle detection.

PR #397, opened 2024-03-06
Implement Tree -> RVSDG Translation

oflatt:
This PR implements the last missing piece in the translation code, converting the Tree IR to an RVSDG.
It also runs all the tests (excluding ones involving memory) on the full compiler.

### DIFF
--- a/src/rvsdg/from_tree.rs
+++ b/src/rvsdg/from_tree.rs
@@ -3,7 +3,9 @@
 use std::iter;
 
 use bril_rs::{ConstOps, EffectOps, Literal, ValueOps};
-use tree_in_context::schema::{BaseType, BinaryOp, Expr, RcExpr, TreeProgram, Type, UnaryOp};
+use tree_in_context::schema::{
+    BaseType, BinaryOp, Expr, Order, RcExpr, TreeProgram, Type, UnaryOp,
+};
 
 use super::{BasicExpr, Operand, RvsdgBody, RvsdgFunction, RvsdgProgram, RvsdgType};
 
@@ -163,6 +165,7 @@ impl<'a> TreeToRvsdg<'a> {
 
     fn convert_expr(&mut self, expr: RcExpr) -> Operands {
         match expr.as_ref() {
+            Expr::Function(..) => panic!("Expected non-function expr in convert_expr"),
             Expr::Const(constant, ty) => match constant {
                 tree_in_context::schema::Constant::Int(integer) => self.push_basic(
                     BasicExpr::Const(ConstOps::Const, Literal::Int(*integer), bril_rs::Type::Int),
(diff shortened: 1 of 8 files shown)


src/rvsdg/from_tree.rs
```diff
@@ -0,0 +1,412 @@
...
+        }
+        Type::TupleT(_) => panic!("Tuple types not supported in RVSDG"),
+        Type::Unknown => panic!("Unknown type in tree_type_to_rvsdg_types"),
+    }
+}
+
+fn convert_func_output_type(ty: Type) -> Option<bril_rs::Type> {
+    match ty {
+        Type::TupleT(inner) => {
+            assert_eq!(
```
ezrosent, line 58: nit: assert!(inner.is_empty()) ? (and maybe add `{inner:?}` to the assertion string?

src/rvsdg/from_tree.rs
```diff
@@ -0,0 +1,412 @@
...
+        }
+        _ => Some(bril_type_from_type(ty)),
+    }
+}
+
+fn tree_func_to_rvsdg(func: RcExpr, program: &TreeProgram) -> RvsdgFunction {
+    let output_type = func.func_output_ty().expect("Expected function types");
+
+    let Type::TupleT(input_types) = func.func_input_ty().expect("Expected function types") else {
+        panic!("Expected tuple type for inputs in tree_func_to_rvsdg")
```
ezrosent, line 73: Add debug printing to this message (what type did we get instead?)

src/rvsdg/from_tree.rs
```diff
@@ -0,0 +1,412 @@
...
+        // initial state edge is last argument
+        current_state_edge: Operand::Arg(input_types.len()),
+        // initial arguments are the first n arguments
+        current_args: (0..input_types.len()).map(Operand::Arg).collect(),
+    };
+
+    let converted = converter.convert_expr(func.clone());
+
+    let resulting_state_edge = converter.current_state_edge;
+    drop(converter);
```
ezrosent, line 92: question: why do we need the drop? Does NLL not handle this for you?

src/rvsdg/from_tree.rs
```diff
@@ -0,0 +1,412 @@
...
+            .into_iter()
+            .map(|ty| RvsdgType::Bril(bril_type_from_type(ty)))
+            .chain(iter::once(RvsdgType::PrintState))
+            .collect(),
+        nodes,
+        // functions return a single value and a state edge
+        // or just a state edge
+        results: match convert_func_output_type(output_type) {
+            Some(func_type) => {
+                assert!(converted.len() == 1, "Expected exactly one result");
```
ezrosent, line 108: nit: assert_eq! ?

ezrosent (approved):
Pretty straightforward!

What's the issue with the memory ops?

oflatt:
Nothing wrong with memory ops, I haven't tested it. That's my next task

src/rvsdg/from_tree.rs
oflatt: We don't need it, I deleted it.
I was confused at some point.

PR #398, opened 2024-03-07
Fix substitution to add correct types to leaf nodes

oflatt:


tree_in_context/src/type_analysis.egg
```diff
@@ -40,39 +40,89 @@
...
-        (HasArgType e ty)
-      )
+(rule ((= lhs (Switch pred (Cons branch rest)))
+       (HasArgType branch ty))
+      ((HasArgType lhs ty))
+      :ruleset type-analysis)
+;; demand with one fewer branches
+(rule ((= lhs (Switch pred (Cons branch rest))))
+      ((Switch pred rest))
+      :ruleset type-analysis)
```
clyben, line 110: Is this rule creating a new switch? Will this create uncessary terms?

tree_in_context/src/type_analysis.egg
```diff
@@ -40,39 +40,89 @@
       ((panic "arg type mismatch"))
       :ruleset type-analysis)
 
-; Pushing expected arg types down
```
clyben, line 43: Why these pushing down rules no longer needed?

clyben (approved):
Looks good to me! Have two questions on type analysis.

tree_in_context/src/type_analysis.egg
oflatt: We are now going bottom-up instead of top-down

PR #404, opened 2024-03-11
Rust context creation

oflatt:
This PR adds a rust function that adds context to the original program added to the egraph. This allows the first version of the program to have full contextual information.
We can then use egglog rules to create contexts we need later.

This also makes the generated egglog file much smaller by sharing common sub-expressions.

### DIFF
--- a/tree_in_context/src/add_context.rs
+++ b/tree_in_context/src/add_context.rs
@@ -1,3 +1,7 @@
+//! Adds context to the tree program.
+//! The `add_context` method recursively adds context to all of the nodes in the tree program
+//! by remembering the most recent context (ex. Let or If).
+
 use crate::{
     ast::{in_context, infunc},
     schema::{Assumption, Expr, RcExpr, TreeProgram},
@@ -64,7 +68,7 @@ impl Expr {
                 let new_case_num = case_num.add_context(current_ctx.clone());
                 let new_branches = branches
                     .iter()
-                    .map(|b| b.clone().add_context(current_ctx.clone()))
+                    .map(|b| b.add_context(current_ctx.clone()))
                     .collect();
                 RcExpr::new(Expr::Switch(new_case_num, new_branches))
             }
@@ -86,8 +90,9 @@ impl Expr {
                 x.add_context(current_ctx.clone()),
                 y.add_context(current_ctx),
             )),
-            // overwrite old context with new, more specific one
-            Expr::InContext(_old_assumption, child) => child.add_context(current_ctx),
+            Expr::InContext(..) => {
+                panic!("add_context expects a term without context")
+            }
             Expr::Function(..) => panic!("Function should have been handled in func_add_context"),
         }
     }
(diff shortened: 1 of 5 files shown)


tree_in_context/src/lib.rs
```diff
@@ -37,11 +41,54 @@ pub fn prologue() -> String {
     .join("\n")
 }
 
+// Returns a variable representing the
```
ezrosent, line 44: doc comment, and trails off

tree_in_context/src/lib.rs
```diff
@@ -37,11 +41,54 @@ pub fn prologue() -> String {
...
+fn print_with_intermediate_helper(
+    termdag: &TermDag,
+    term: Term,
+    cache: &mut HashMap<Term, String>,
+    res: &mut String,
```
ezrosent, line 49: can you implement this as an `impl io::Write` instead? Instead of `x.push_str(format!(...))` you can do `write!(x, ...)`, and you also avoid a copy.

tree_in_context/src/add_context.rs
```diff
@@ -0,0 +1,94 @@
+use crate::{
```
ezrosent, line 1: Add a module doc?

tree_in_context/src/add_context.rs
```diff
@@ -0,0 +1,94 @@
...
+                    new_pred,
+                    then_case.add_context(then_ctx),
+                    else_calse.add_context(else_ctx),
+                ))
+            }
+            Expr::Switch(case_num, branches) => {
+                let new_case_num = case_num.add_context(current_ctx.clone());
+                let new_branches = branches
+                    .iter()
+                    .map(|b| b.clone().add_context(current_ctx.clone()))
```
ezrosent, line 67: nit: do you need this clone? It looks like `add_context` is effectively doing a clone

tree_in_context/src/add_context.rs
```diff
@@ -0,0 +1,94 @@
...
+                RcExpr::new(Expr::Call(f.clone(), arg.add_context(current_ctx.clone())))
+            }
+            Expr::Single(e) => RcExpr::new(Expr::Single(e.add_context(current_ctx))),
+            Expr::Concat(order, x, y) => RcExpr::new(Expr::Concat(
+                order.clone(),
+                x.add_context(current_ctx.clone()),
+                y.add_context(current_ctx),
+            )),
+            // overwrite old context with new, more specific one
+            Expr::InContext(_old_assumption, child) => child.add_context(current_ctx),
```
ezrosent, line 90: This comment is a bit mysterious:

* When can we have two overlapping InContext nodes?
* Why are later visits more specific?

ezrosent (approved):
minor comments, and a question.

tree_in_context/src/add_context.rs
oflatt: Actually we don't call this method in any places where there is context, so I just made this case panic instead
But the idea was that this pass adds the most specific possible context, since it includes the entire program path.

PR #406, opened 2024-03-14
Context rules with fine-grained control of how deep to copy

oflatt:
This PR improves the context rules, with fine-grained control of how deep to copy.
It also fixes our saturation issues by keeping track of a list of eclasses already visited, and subsuming rows that have encountered a cycle.

### DIFF
--- a/tree_in_context/src/utility/in_context.egg
+++ b/tree_in_context/src/utility/in_context.egg
@@ -17,6 +17,16 @@
   (PathNil)
   (PathCons Expr Expr ContextPath))
 
+
+;; For every path P in the database,
+;; we add a (Path-contains P S)
+;; for every subpath S with the same endpoint.
+;; We also include the empty path.
+;; Ex: for path P = (PathCons a (PathCons b (PathCons c (PathNil))))
+;; we have (Path-contains P P),
+;; (Path-contains P (PathCons b (PathCons c (PathNil)))),
+;; (Path-contains P (PathCons c (PathNil))),
+;; and (Path-contains P (PathNil)).
 (relation Path-contains (ContextPath ContextPath))
 
 (rule ((= lhs (PathCons hd1 hd2 tl)))
@@ -213,7 +223,4 @@
          (DoWhile new-inputs
            (DoAddContext newpath (InLoop new-inputs outputs) scope outputs))))
        :ruleset context)
-
-
-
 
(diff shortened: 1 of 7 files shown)


tree_in_context/src/utility/in_context.egg
```diff
@@ -1,146 +1,219 @@
...
+       (= scope (Full)))
+      ((let new-inputs
+         (DoAddContext seen ctx scope inputs))
+       (let newpath
+         (PathCons (DoWhile inputs outputs) lhs seen))
+       (union lhs
+         (DoWhile new-inputs
+           (DoAddContext newpath (InLoop new-inputs outputs) scope outputs))))
+       :ruleset context)
+
```
ezrosent, line 216: Probably don't need all these extra lines

tree_in_context/src/utility/in_context.egg
```diff
@@ -1,146 +1,219 @@
...
+         (Single (DoAddContext seen ctx scope c1))
+         :ruleset context-helpers)
+(rewrite (DoAddContext seen ctx scope (Concat order c1 c2))
+         (Concat order
+           (DoAddContext seen ctx scope c1)
+           (DoAddContext seen ctx scope c2))
+         :ruleset context-helpers)
+
+;; Replace existing contexts.
+;; This relies on the fact that new contexts must be more specific than old ones.
```
ezrosent, line 158: I know we discussed this in a meeting at some point, but have we written down why this is true anywhere (and continues to be true, say, when we add some of these contexts ahead of time in the Rust pass?)

ezrosent (approved):

tree_in_context/src/utility/in_context.egg
oflatt: That's a great question, and I hope to answer that soon by updating the semantics in the repository.
A high-level answer is that we maintain this invariant:
A context node defines the set of values that a program is valid on, and this program is *guaranteed to execute* on these values. That means nested context nodes must pass for all values the program is executed on.

tree_in_context/src/utility/in_context.egg
```diff
@@ -1,146 +1,219 @@
...
+;; A context path is a list of (eclass, eclass) pairs.
+;; In substitution, the path is used to store every context-related eclass we have visited, 
+;; and the corresponding expression at that point.
+;; When we visit an eclass for the second time, we have found a cycle in the egraph.
+;; Then the corresponding eclasses are made equivalent.
+(datatype ContextPath
+  (PathNil)
+  (PathCons Expr Expr ContextPath))
+
+(relation Path-contains (ContextPath ContextPath))
```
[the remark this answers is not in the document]
oflatt: added a description with example

PR #408, opened 2024-03-24
Brilift integration

### LINKED ISSUE #385: Add `brilift` integration to compile Bril to LLVM
https://capra.cs.cornell.edu/bril/tools/brilift.html

https://github.com/sampsyo/bril/tree/main/brilift

Alex-Fischman:
Adds `RunType::Compile` and `Interpretable::Executable` to `util.rs`, which calls into `brilift` to compile and run the program.

Resolves #385.


src/util.rs
```diff
@@ -522,6 +527,43 @@ impl Run {
...
+
+                let executable = std::path::Path::new(&executable).canonicalize().unwrap();
+
+                std::process::Command::new("rm")
+                    .arg(object)
+                    .status()
+                    .unwrap();
+
+                // TODO: only do this during tests
+                if !self.interp {
```
oflatt, line 558: We should probably produce temporary files and directories for tests
Here's a crate for that: https://docs.rs/tempdir/latest/tempdir/

infra/brilift.c
```diff
@@ -0,0 +1,67 @@
+#include <stdio.h>
```
oflatt, line 1: Is this copied from somewhere? We should at least link to where it came from.
Perhaps the crate could expose it somehow, as a string or file? It may get out of sync at some point

oflatt (changes requested):
Awesome work! Had some minor questions

infra/brilift.c
Alex-Fischman: It's copied from [https://github.com/sampsyo/bril/blob/main/brilift/rt.c]. I don't know how to keep it synced though.

oflatt: Could brillift have a function that spits out the string using the `include_str` macro?

Alex-Fischman: https://github.com/sampsyo/bril/pull/314

oflatt (approved):
great work!

PR #410, opened 2024-03-27
DAG with Context IR

oflatt:
This is the working branch for transitioning to a DAG-based IR. Make PRs into this branch.


oflatt:
I'm no longer committing to this branch. Instead, we'll make PRs to it.

PR #411, opened 2024-03-27
Interpreter for DAG IR

oflatt:
This PR fixes up the interpreter to work with the DAG IR. It does so by caching results for evaluating nodes based on the Rc pointer so that they are evaluated just once.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -67,8 +67,7 @@ impl Optimizer {
     }
 
     /// Produces a tuple value from the string arguments
-    fn parse_arguments(args: Vec<String>) -> Value {
-        Value::Tuple(
+    fn parse_arguments(args: Vec<String>) -> Vec<Value> {
             args.into_iter()
                 .map(|arg| {
                     if let Ok(int) = arg.parse::<i64>() {
@@ -81,8 +80,7 @@ impl Optimizer {
                         panic!("Invalid argument to bril program: {}", arg);
                     }
                 })
-                .collect(),
-        )
+                .collect()
     }
 
     /// Interpret a program in an `Interpretable` IR.
@@ -96,8 +94,11 @@ impl Optimizer {
         match program {
             Interpretable::Bril(program) => Self::interp_bril(program, args, profile_out),
             Interpretable::TreeProgram(program) => {
-                let (val, mut printed) = interpret_tree_prog(program, &Self::parse_arguments(args));
-                assert_eq!(val, Value::Tuple(vec![]));
+                let mut parsed = Self::parse_arguments(args);
+                // add the state value to the end
+                parsed.push(Value::StateV);
+                let (val, mut printed) = interpret_tree_prog(program, &Value::Tuple(parsed));
+                assert_eq!(val, Value::Tuple(vec![Value::StateV]));
                 // add new line to the end of each line in printed
                 for line in printed.iter_mut() {
                     line.push('\n');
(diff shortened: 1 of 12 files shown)


src/lib.rs
```diff
@@ -67,22 +67,20 @@ impl Optimizer {
     }
 
     /// Produces a tuple value from the string arguments
```
ajpal, line 69: nit: update comment

tree_in_context/src/interpreter.rs
```diff
@@ -1,6 +1,14 @@
-use std::{collections::HashMap, fmt::Display};
+//! Interpret DAG programs. Each expression is evaluated once, so side-effects only happen
+//! once. All the dependencies of an expression are evaluted first.
+//! The interpreter relies on the invariant that common subexpressions are
```
ajpal, line 3: Is this invariant checked/enforced or is this comment the only place it's clearly documented?

tree_in_context/src/interpreter.rs
```diff
@@ -83,6 +98,8 @@ pub(crate) struct VirtualMachine<'a> {
     next_addr: usize,
     /// All of memory
     mem: HashMap<usize, Value>,
+    /// Values for already evaluated expressions
+    memo: HashMap<*const Expr, Value>,
```
ajpal, line 102: nit: `mem` and `memo` are quite similar, is there a better name for one?

tree_in_context/src/interpreter.rs
```diff
@@ -263,11 +310,14 @@ impl<'a> VirtualMachine<'a> {
...
-            Expr::Alloc(e_size, _ignore_state, _ty) => {
+            Expr::Alloc(e_size, state_expr, _ty) => {
                 let size = self.interp_int_expr(e_size, arg);
+                let state_val = self.interpret_expr(state_expr, arg);
+                assert_eq!(state_val, Value::StateV);
                 let addr = self.next_addr;
                 self.next_addr += usize::try_from(size).unwrap();
-                Ptr(Pointer::new(addr, size as usize, 0))
+
+                tuplev!(Ptr(Pointer::new(addr, size as usize, 0)), Value::StateV)
```
ajpal, line 320: what's the 0 for?

tree_in_context/src/typechecker.rs
```diff
@@ -51,19 +51,25 @@ impl Expr {
     }
 }
 
+/// A map from the old, untyped expression to the new, typed expression
+pub type TypedExprCache = HashMap<*const Expr, RcExpr>;
```
ajpal, line 55: what's the difference between TypeExprCache and TypeCache?

tree_in_context/src/interpreter.rs
oflatt: 0 is the offset for the pointer (alloc returns a new memory region with a pointer pointing to the start of it)
Added a comment

tree_in_context/src/typechecker.rs
oflatt: Good question, I wrote new documentation to try to clarify
```
/// Typechecking produces new, typed expressions.
/// This map is used to memoize the results of typechecking.
/// It maps the old untyped expression to the new typed expression
pub type TypedExprCache = HashMap<*const Expr, RcExpr>;

/// We also need to keep track of the type of the newly typed expression.
/// This maps the newly instrumented expression to its type.
pub type TypeCache = HashMap<*const Expr, Type>;
```

ajpal (approved):

PR #413, opened 2024-04-01
Dag to RVSDG conversion algorithm

oflatt:
This PR adds an algorithm to transform the DAG IR back to RVSDG. The hard part is finding what region if statements have.


src/rvsdg/dag_to_graph.rs
```diff
@@ -0,0 +1,276 @@
...
+                rgraph
+                    .expr_branch_node
+                    .insert((Rc::as_ptr(&expr), 0), then_branch_node);
+                let else_branch_node = rgraph.graph.add_node(());
+                rgraph
+                    .expr_branch_node
+                    .insert((Rc::as_ptr(&expr), 1), else_branch_node);
+                let expr_node = rgraph.node(&expr);
+                let inputs_node = rgraph.node(inputs);
+                let then_node = rgraph.node(then_branch);
```
ajpal, line 62: I don't think I understand the difference btween `then_node` and `then_branch_node` (and same for else)?

ajpal (approved):
Left a few questions, but overall looks good! great work on this :)

PR #416, opened 2024-04-02
Add a DAG to SVG converter for visualizing the DAG IR

oflatt:
This PR adds a visualizer for the DAG IR
It also removes a snapshot test that consistently causes us problems and isn't useful


Alex-Fischman (changes requested):
LGTM, except for all the places in the code that say Tree instead of Dag. IDK what's actually a tree vs. a DAG anymore, but at the very least TreeProgram seems wrong.

oflatt:
more name changes to come

PR #419, opened 2024-04-03
Extraction on serialized E-graph from eggcc

yihozhang:
future TODOs: Ignore children nodes sharing for region nodes like Loops.


oflatt (approved):
great!

PR #421, opened 2024-04-04
Add context before sending to egglog

oflatt:
This PR makes the optimizer add context nodes before sending the program to egglog.
To ensure linearity of the state edge on the way back, it removes these context nodes again. This is a hack that should be fixed by a linearity-preserving extractor in the future. Right now, eggcc assumes the extracted program uses memory linearly.

### DIFF
--- a/dag_in_context/src/from_egglog.rs
+++ b/dag_in_context/src/from_egglog.rs
@@ -314,7 +314,7 @@ impl FromEgglog {
         res
     }
 
-    pub fn program_from_egglog(&mut self, program: Term) -> TreeProgram {
+    fn program_from_egglog_preserve_ctx_nodes(&mut self, program: Term) -> TreeProgram {
         match_term_app!(program.clone();
         {
           ("Program", [entry, functions]) => {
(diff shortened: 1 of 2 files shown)


oflatt:
err idk why add if to CICM is in the diff

dag_in_context/src/from_egglog.rs
```diff
@@ -327,4 +327,33 @@ impl FromEgglog {
...
+    }
+
+    fn without_ctx_nodes(&mut self, expr: Term) -> Term {
+        match expr {
+            Term::App(head, children) => match (head.to_string().as_str(), children.as_slice()) {
+                ("InContext", [_assumption, expr]) => {
+                    self.without_ctx_nodes(self.termdag.get(*expr))
+                }
+                _ => {
+                    let mut new_children = vec![];
```
rtjoa, line 347: nit: iter and collect might be easier to read

rtjoa (approved):
Looks good. A small unit test would be nice

dag_in_context/src/from_egglog.rs
oflatt: had lifetime issues

PR #422, opened 2024-04-04
Bril LLVM

ryan-berger:


oflatt:
Marking as draft until it runs tests

src/main.rs
```diff
@@ -75,6 +75,7 @@ fn main() {
         output_path: args.output_path,
         optimize_egglog: args.optimize_egglog,
         optimize_brilift: args.optimize_brilift,
+        optimize_brillvm: args.optimize_brilift,
```
oflatt, line 78: bug here? is the brillvm flag supposed to be different?

tests/files.rs
```diff
@@ -26,6 +26,14 @@ fn generate_tests(glob: &str) -> Vec<Trial> {
                     .unwrap();
             }
 
+            if run.test_type == RunType::CompileBrilLLVM {
```
oflatt, line 29: this should probably be the same code path as the if above
you should also generate an args file like the compileBrilift mode for use in nightlies

oflatt (changes requested):
Great work! Some minor things

src/main.rs
ryan-berger: Yes, definitely a bug, nice catch

oflatt (approved):
Thanks!

PR #424, opened 2024-04-05
Add ids to Allocs

rtjoa:
Add an `i64` field to `Alloc`, which should be uniquely assigned during translation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -599,7 +599,7 @@ dependencies = [
 [[package]]
 name = "egglog"
 version = "0.1.0"
-source = "git+https://github.com/rtjoa/egglog?rev=77ffa1f#77ffa1f7599b5e2d8936a5b78fa89bfa563bf2bc"
+source = "git+https://github.com/rtjoa/egglog?rev=e3a3ae#e3a3ae76e8caafc55d229f7d4f1ce7a095240130"
 dependencies = [
  "clap",
  "egraph-serialize",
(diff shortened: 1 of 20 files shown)


oflatt:
I think that we should make the allocations have unique ids in rust, not egglog
That way we don't need `i64-fresh!`

I'd also be a lot happier if we found a way to not have unique ids on the allocations, and use the expression that generated that allocation instead. But we'd need to think carefully about that

rtjoa:
> I think that we should make the allocations have unique ids in rust, not egglog That way we don't need `i64-fresh!`

I originally did it from rust by adding a `next_alloc_id: i64` field to `DagTranslator`. However, we then have no way to atomically get the next id in egglog, which we need to do as we create `Alloc`s in `Subst`/`DoAddContext`.

> I'd also be a lot happier if we found a way to not have unique ids on the allocations, and use the expression that generated that allocation instead. But we'd need to think carefully about that

We could refer to allocs in a *child* context or a func call with this cell datatype:
```
type cell = InSameContext expr | InLoop in out expr | InFuncCall fn args expr
```
but we'd need another Context node type to refer to allocs in a parent context.

For now, I think unique ids is easiest our current plan as of last memory meeting.

rtjoa:
Per discussion on Slack, assigned ids in Rust, as it's actually sound to have multiple ids per e-class: it just makes the analysis less precise.

oflatt (approved):
lgtm

PR #425, opened 2024-04-05
Update Alloc to take the type it allocates instead of pointer type it creates

rtjoa:


oflatt:
closing for now to clean up the PR list, not sure if you are working on this

PR #417, opened 2024-04-02
[Nightly] Benchmark brilift executables

ajpal:
Still to do:
- add other run-modes (eggcc + brilift, eggcc only, brillvm)
- clean up the UI

PR #426, opened 2024-04-05
[Nightly] Pass args through to executable

ajpal:
Follow up to https://github.com/egraphs-good/eggcc/pull/417, fixes the segfaults by passing args through


src/util.rs
```diff
@@ -660,11 +667,17 @@ impl Run {
...
             .unwrap();
 
-        if self.in_test && !self.interp {
+        if self.in_test {
             std::process::Command::new("rm")
-                .arg(executable.clone())
+                .arg(executable.clone() + "-args")
                 .status()
                 .unwrap();
+            if !self.interp {
```
oflatt, line 675: why do we only remove the executable when not interpreting?
The tests should always clean up executables.

infra/profile.sh
```diff
@@ -27,11 +27,13 @@ bench() {
...
   cargo run --release "$1" --run-mode compile-brilift -o "$PROFILE_DIR/brilift"
-  
-  # TODO: Some of the brils result in executables that seg fault, need to figure out why
-  # For now, though, we just silently ignore them
-  hyperfine --warmup 2 --export-json "$PROFILE_DIR"/brilift.json "$PROFILE_DIR/brilift" || echo "[BRILIFT] could not run $PROFILE_NAME"
+
+  # Read the args from the file
+  ARGS=$(cat $PROFILE_DIR/brilift-args)
+
+  hyperfine --warmup 2 --export-json $PROFILE_DIR/brilift.json "$PROFILE_DIR/brilift $ARGS"
```
oflatt, line 36: You'll probably want to clean up the executable and argument files here
I'd say that all of these should go in a temporary directory, not the top level

oflatt (changes requested):
A couple suggestions with managing files

src/util.rs
ajpal: I'm maintaining existing behavior here: https://github.com/egraphs-good/eggcc/blob/main/src/util.rs#L663

infra/profile.sh
ajpal: The profiler does put files in a tmp directory and clean them up itself (see lines 5-9 of this file)

oflatt (approved):
lgtm

PR #429, opened 2024-04-05
Improve DAG SVG renderer with proper regions

oflatt:


oflatt:
closing but will re-open with cleaner hist

PR #430, opened 2024-04-05
Improve dag visualizer to properly create regions

oflatt:


dag_in_context/src/dag2svg.rs
```diff
@@ -3,24 +3,69 @@ use std::{
...
-use crate::schema::{Expr, RcExpr, TreeProgram};
+use crate::schema::{Constant, Expr, RcExpr, TreeProgram};
+
+// We need one node for every unique expression, scope pair
+// Expressions may be shared across scopes but should be shown multiple times
+#[derive(Eq, PartialEq, Hash)]
+struct UniqueExpr {
+    expr: *const Expr,
+    scope: *const Expr,
+}
```
Alex-Fischman, line 16: Why are you using `*const Expr` instead of `&Expr`? This is very odd, isn't it?

dag_in_context/src/dag2svg.rs
```diff
@@ -3,24 +3,69 @@ use std::{
...
+        if let Some(name) = self
+            .get_name
+            .get(&UniqueExpr::new(self.current_scope, expr))
+        {
             Id::Plain(name.clone())
         } else {
-            let name = format!("{}{}", expr.constructor().name(), self.name_counter);
+            let name = match expr.as_ref() {
+                Expr::DoWhile(_, _) => {
+                    format!("cluster_{}{}", expr.constructor().name(), self.name_counter)
```
Alex-Fischman, line 44: Why `cluster` instead of `DoWhile`?

dag_in_context/src/dag2svg.rs
```diff
@@ -76,22 +129,50 @@ impl Expr {
...
                     stmts: body.to_dot_with(conv),
-                    id: conv.graphviz_id(self),
+                    id: id_outside,
+                }));
+                stmts.push(Stmt::Edge(Edge {
+                    ty: EdgeTy::Pair(vertex_outside, conv.graphviz_vertex(body)),
+                    attributes: vec![],
                 }));
+                // restore scope of the dowhile
+                conv.current_scope = scope_before;
```
Alex-Fischman, line 156: Why is this in `DoWhile` but not below in `Function`?

dag_in_context/src/dag2svg.rs
oflatt: It's a dot thing: naming it cluster tells dot to create a group of nodes inside a box

dag_in_context/src/dag2svg.rs
oflatt: `Function` is a top-level scope so there's no scope above it

dag_in_context/src/dag2svg.rs
oflatt: added comment

dag_in_context/src/dag2svg.rs
oflatt: added comment

dag_in_context/src/dag2svg.rs
oflatt: added comment

PR #431, opened 2024-04-05
Fix bug in sharing invariant via translation to Term and back again

oflatt:
The RVSDG->DAG pass wasn't maintaining the sharing invariant (that sub-terms that are equal must be shared using the same Rc<Expr> poitner).
This PR fixes that.

It also fixes up typechecking to do proper caching based on the argument type


Alex-Fischman (approved):
LGTM but I think this has overlap with your other PRs?

oflatt:
I think just the benchmark overlaps

PR #435, opened 2024-04-05
Switch back to main bril

ajpal:
Now that https://github.com/sampsyo/bril/pull/315 is merged, we shouldn't depend on my fork anymore.


oflatt (approved):
thanks!

PR #436, opened 2024-04-06
ContextOf (propagates contexts up from leaves)

kirstenmg:
Like ExprHasRefId or IdOf, ContextOf propagates context information up from leaves in the current context, so we know the context of an expression. We do not need a ListExpr version of ContextOf, as the only ListExpr child in the schema is of Switch; Switch takes its context from its predicate rather than its branches.

This PR contains some commented-out code that checks that one expression doesn't get two nonequivalent contexts, since it is not certain whether that's a problem.


dag_in_context/src/utility/context_of.egg
```diff
@@ -0,0 +1,65 @@
+; We only have context for Exprs, not ListExprs.
+(relation ContextOf (Expr Assumption))
+
+; Base case: InContext means ContextOf
+(rule ((InContext ctx e))
+    ((ContextOf (InContext ctx e) ctx)) :ruleset always-run)
+
+; Error checking - each expr should only have a single context
+; TODO: this may not be true
```
clyben, line 9: I think this is not true considering common subexpr?

dag_in_context/src/utility/context_of.rs
```diff
@@ -0,0 +1,93 @@
...
+    let pred = eq(arg(), int(5)).with_arg_types(base(intt()), base(boolt()));
+    let body = tif(pred.clone(), arg(), int(4)).with_arg_types(base(intt()), base(intt()));
+    let build = function("main", base(intt()), base(intt()), body.clone()).func_add_context();
+
+    // If statement should have the context of its predicate
+    let check = "
+        (let pred-ctx (InFunc \"main\"))
+        (let pred (Bop (Eq) (InContext (InFunc \"main\") (Arg (Base (IntT)))) (InContext (InFunc \"main\") (Const (Int 5) (Base (IntT))))))
+        (check (ContextOf pred pred-ctx))
+        (let if (If (Bop (Eq) (InContext (InFunc \"main\") (Arg (Base (IntT)))) (InContext (InFunc \"main\") (Const (Int 5) (Base (IntT))))) (InContext (InIf true (Bop (Eq) (InContext (InFunc \"main\") (Arg (Base (IntT)))) (InContext (InFunc \"main\") (Const (Int 5) (Base (IntT)))))) (Arg (Base (IntT)))) (InContext (InIf false (Bop (Eq) (InContext (InFunc \"main\") (Arg (Base (IntT)))) (InContext (InFunc \"main\") (Const (Int 5) (Base (IntT)))))) (Const (Int 4) (Base (IntT))))))
```
clyben, line 15: Could you pretty print these lines to
```
(let pred-ctx (InFunc \"main\"))
(let pred (Bop (Eq) (InContext pred-ctx (Arg (Base (IntT)))) (InContext pred-ctx (Const (Int 5) (Base (IntT))))))
(check (ContextOf pred pred-ctx))
(let if 
      (If pred (InContext (InIf true (Bop (Eq) (InContext pred-ctx (Arg (Base (IntT)))) (InContext pred-ctx (Const (Int 5) (Base (IntT)))))) (Arg (Base (IntT)))) 
               (InContext (InIf false (Bop (Eq) (InContext pred-ctx (Arg (Base (IntT)))) (InContext pred-ctx (Const (Int 5) (Base (IntT)))))) (Const (Int 4) (Base (IntT))))))
(check (ContextOf if pred-ctx))
```

clyben (changes requested):
Great thanks! Though with some questions. 
I think `(InContext (InFun a) (Num 1)` and `(InContext (InFun b) (Num 1)` can exist at the same time so an Expr can have different context?

dag_in_context/src/utility/context_of.egg
kirstenmg: It may not be true, but not because of common subexpressions. `ContextOf` only contains expressions that have `InContext` nodes within them. For example, consider your example of `(InContext (InFun a) (Num 1))` and `(InContext (InFun b) (Num 1))`. It is not true that `(ContextOf (Num 1) (InFun a))`, since just `(Num 1)` is contextless. However, we do know:
`(ContextOf (InContext (InFun b) (Num 1)) (InFun b))`
`(ContextOf (InContext (InFun a) (Num 1)) (InFun b))`
`(ContextOf (Add (InContext (InFun a) (Num 1)) (InContext (InFun a) (Num 1))))`
and so on.

What should not happen is something like this, where contexts that are not equivalent are mixed:
`(ContextOf (Add (InContext (InFun a) (Num 1)) (InContext (InFun b) (Num 1))))`
That is the motivation for this rule. However, I believe that there are cases where two contexts that have not yet been discovered to be equal are mixed? I am not exactly sure on why two contexts is OK sometimes, Oliver and Ryan were discussing this.

dag_in_context/src/utility/context_of.egg
```diff
@@ -0,0 +1,65 @@
+; We only have context for Exprs, not ListExprs.
```
oflatt, line 1: remove underscores now that we aren't generating

dag_in_context/src/utility/context_of.rs
```diff
@@ -0,0 +1,96 @@
...
+        (let pred (Bop (Eq) (InContext (InFunc \"main\") (Arg (Base (IntT)))) (InContext (InFunc \"main\") (Const (Int 5) (Base (IntT))))))
+        (check (ContextOf pred pred-ctx))
+        (let if
+                (If pred
+                    (InContext (InIf true (Bop (Eq) (InContext (InFunc \"main\") (Arg (Base (IntT)))) (InContext (InFunc \"main\") (Const (Int 5) (Base (IntT)))))) (Arg (Base (IntT))))
+                    (InContext (InIf false (Bop (Eq) (InContext (InFunc \"main\") (Arg (Base (IntT)))) (InContext (InFunc \"main\") (Const (Int 5) (Base (IntT)))))) (Const (Int 4) (Base (IntT))))))
+        (check (ContextOf if pred-ctx))
+        ".to_string();
+
+    // Don't pass in any programs, just want build/check to be checked
```
oflatt, line 22: we should always interpret our programs! it helps catch bugs

oflatt (changes requested):
Some minor changes

oflatt (approved):
yay!

PR #438, opened 2024-04-09
add workaround for new macos linker

ezrosent:
All the tests pass on my machine.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -671,7 +671,7 @@ impl Run {
             //
             // This is either a bug, or a difference in the way symbols are
             // handled, or a bit of both (chatter online differs), but for now,
-            // we just rety with the ld_classic flag.
+            // we just retry with the ld_classic flag.
             if !cmd
                 .stderr(Stdio::null())
                 .status()
--- a/src/util.rs
+++ b/src/util.rs
@@ -11,6 +11,7 @@ use graphviz_rust::printer::PrinterContext;
 use std::fmt::Debug;
 use std::fs;
 use std::io::Read;
+use std::process::Stdio;
 use std::{
     ffi::OsStr,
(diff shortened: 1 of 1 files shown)


Alex-Fischman:
Works locally for me.

src/util.rs
```diff
@@ -655,13 +656,47 @@ impl Run {
...
+        {
+            // Workaround on new macos linkers:
+            //
+            // On linkers shipped past XCode 15, we see a bug around symbol
+            // relocations with an error along the lines of:
+            // ld: Assertion failed: (pattern[0].addrMode == addr_other), function addFixupFromRelocations, file Relocations.cpp, line 701.
+            //
+            // This is either a bug, or a difference in the way symbols are
+            // handled, or a bit of both (chatter online differs), but for now,
+            // we just rety with the ld_classic flag.
```
Alex-Fischman, line 669: nit: typo

src/util.rs
```diff
@@ -655,13 +656,47 @@ impl Run {
...
+        #[cfg(not(target_os = "macos"))]
+        {
+            std::process::Command::new("cc")
+                .arg(object.clone())
+                .arg(library_o.clone())
+                .arg("-o")
+                .arg(executable.clone())
+                .status()
+                .unwrap();
+        }
```
Alex-Fischman, line 699: nit: can you factor out the base command
```
std::process::Command::new("cc")
                .arg(object.clone())
                .arg(library_o.clone())
                .arg("-o")
                .arg(executable.clone())
```
outside of the `cfg` blocks? I think it'll be much easier to read if this is built in a local var first, and then modified/called three times.

Alex-Fischman (approved):
Couple of nits, once they're fixed I don't need to re-review

PR #440, opened 2024-04-09
Fix parallel to nest like a linked list

Alex-Fischman:
This makes the `parallel!` macro nest to the right and adds a canonicalization rule to the optimizer.


oflatt (approved):
nice

PR #441, opened 2024-04-09
Add context during RVSDG->DAG translation, avoiding code size blowup

oflatt:
Adding contexts can exponentially blow up code size now that conditionals do not have regions. 
This PR fixes that problem on the rust side by only adding context up to RVSDG region boundaries.
We'll have to be much more careful on the egglog side- this will be fixed in a follow on PR.

This PR also fixes the typechecker to type check context nodes, now that typechecking happens after the translation.


dag_in_context/src/schema.egg
```diff
@@ -118,17 +118,10 @@
 (datatype Order
   ; Either argument may be evaluated first.
   ; Either way, the program should reach an equivalent state.
-  (Parallel)
-  ; Evaluate the first argument first.
-  (Sequential)
-  ; Evaluate the second argument first.
-  (Reversed))
+  (Parallel))
```
ajpal, line 121: If order is always parallel, why do we need it?

dag_in_context/src/typechecker.rs
```diff
@@ -1,11 +1,33 @@
...
 };
 
+pub(crate) struct TypeStack(Vec<Type>);
+
+impl TypeStack {
+    pub(crate) fn get(&self) -> &Type {
+        self.0.last().unwrap()
+    }
+
+    pub(crate) fn pushed(&self, ty: Type) -> TypeStack {
```
ajpal, line 18: why pushed and popped instead of push and pop?

ajpal (approved):
I didn't do a super thorough review, but from cursory glance, lgtm (aside from a few nits)

dag_in_context/src/typechecker.rs
oflatt: they don't mutate self

PR #443, opened 2024-04-09
Remove tuple-ith, just use `Get`

oflatt:
This PR removes tuple-ith, since with linearity issues resolved (soon) we can use Get instead.
This revealed an issue in saturating context rules, but only in tests. It turns out that adding infinite nestings of contexts, even if they are the same one, can result in the context rules not saturating (InLoop refers to a body, which may already have context).


dag_in_context/src/lib.rs
```diff
@@ -211,6 +211,7 @@ pub fn egglog_test(
         prologue(),
         include_str!("schedule.egg"),
     );
+    eprintln!("{}", program);
```
Alex-Fischman, line 214: I think that printing the whole program is bad, both here and on line 225, because if multiple tests fail it pushes the actual errors off the screen, and also scrolling through the terminal to read egglog sucks. Can we maybe just add (or advertise if it already exists) a top-level command to dump the program, and remove these two prints?

dag_in_context/src/schedule.egg
```diff
@@ -1,4 +1,5 @@
 (run-schedule
+  test-helpers
```
Alex-Fischman, line 2: nit: test-helpers is definitely a terrible name since it's not like the ruleset turns off for release

dag_in_context/src/utility/in_context.egg
```diff
@@ -82,23 +86,6 @@
...
-;; (DoWhile inputs (InContext (InLoop inputs ctx1) ctx1))
-;; These two contexts are equal, since the set of reachable
-;; values is the same.
-(rule ((= body_with_context
-          (DoAddContext seen  (InLoop inputs outputs) (Full) outputs)) ;; a loop body with context
-          (DoAddContext seen2 (InLoop inputs body_with_context) ;; same inputs
-                              (Full) body_with_context) ;; a loop body whose context already has context
-       )
-      ((union (InLoop inputs outputs) (InLoop inputs body_with_context)))
-      :ruleset context-helpers)
```
Alex-Fischman, line 100: Why'd we delete this?

dag_in_context/src/utility/util.egg
```diff
@@ -19,22 +19,43 @@
...
+      ((union (Get (Concat expr1 expr2) i)
+              (Get expr1 i)))
+       :ruleset always-run)
+;; descend right
+(rule ((Get (Concat expr1 expr2) i)
+       (= (tuple-length expr1) len1)
+       (>= i len1))
+      ((union (Get (Concat expr1 expr2) i)
+              (Get expr2 (- i len1))))
+       :ruleset always-run)
```
Alex-Fischman, line 57: Why did the rules here get more complicated? Did you just reformat them or is there a semantic difference?

Alex-Fischman (commented):
It feels sus to me to just make the rule run less often, but maybe it's okay.

dag_in_context/src/lib.rs
oflatt: Good catch on this print, committed by mistake
Idk how to add a top-level command to print the program for rust tests. Would add if I knew how

dag_in_context/src/utility/util.egg
oflatt: There is a semantic difference- they are more complete now

dag_in_context/src/schedule.egg
oflatt: suggestions?

oflatt:
Yeah
Ensuring saturation when you can keep making new contexts each iterations has been really painful, I'd like to return to the problem sometime

dag_in_context/src/utility/util.egg
oflatt: More discussion- specifically, they are more complete when the leaf node of the concat isn't a Single. For example, a DoWhile could be concat with something

Alex-Fischman (approved):

PR #445, opened 2024-04-10
[Interval Analysis] Working examples of using context nodes for interval analysis

ajpal:
This PR changes to using separate `lo-bound` and `hi-bound` tables instead of one `ival` table (as discussed in lab today)

The first test roughly corresponds to:
```
if (input <= 0) {
  y = -1 * input
} else {
  y = input
}
return y < 0
```
And the interval analysis reduces this to just `false`

The second test roughly corresponds to:
```
if (0 <= input) {
  y = -1 * input
} else {
  y = input
}
return 0 < y
```
Which is also `false`.

### DIFF
--- a/dag_in_context/src/interval_analysis.egg
+++ b/dag_in_context/src/interval_analysis.egg
@@ -83,6 +83,16 @@
       )
       :ruleset interval-analysis)
 
+; negative number * negative number is positive
+(rule (
+        (= lhs (Bop (Mul) x y))
+        (= (IntB hi-x) (hi-bound x))
+        (= (IntB hi-y) (hi-bound y))
+        (<= hi-x 0)
+        (<= hi-y 0) 
+      )
+      ((set (lo-bound lhs) (IntB 0)))
+      :ruleset interval-analysis)
 
 ; < a b interval is (< ha lb) (< la hb)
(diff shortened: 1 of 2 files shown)


dag_in_context/src/interval_analysis.egg
```diff
@@ -1,87 +1,233 @@
...
+       (set (hi-bound lhs) (IntB (+ ha hb)))
       )
-      ((set (ival lhs) (IntI (+ la hb) (+ ha hb))))
+      :ruleset interval-analysis)
+
+; mul commutes
+(rewrite (Bop (Mul) x y) (Bop (Mul) y x) :ruleset interval-analysis)
+
+; negative number * negative number is positive
+(rule (
```
oflatt, line 106: Seems like you are missing some Mul rules- perhaps in the next PR?
These handle pretty specific cases. The intervals may also be mixed sign (not strictly positive or negative)

oflatt (approved):
Nice work!
The rules are pretty specialized to your example for now, but I'm sure we'll make them more general in the future.

oflatt:
This one will need to be rebased on main (sorry)
Should be pretty easy to migrate, lmk if you have problems

PR #446, opened 2024-04-10
Unshare regions in Dag extraction

yihozhang:


dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -120,9 +120,12 @@ fn get_node_cost(
             .sum::<NotNan<f64>>();
 
     let mut costs: HashMap<ClassId, Cost> = Default::default();
-    for c in child_cost_sets.iter().map(|cs| &cs.costs) {
-        for (cid, cost) in c.iter() {
-            costs.insert(cid.clone(), *cost);
+    // children of region nodes are not shared
+    if !cm.regions.contains(op) {
```
oflatt, line 124: This doesn't seem right for loops- we still want to consider the inputs of the loops in the same region!

oflatt (changes requested):

dag_in_context/src/greedy_dag_extractor.rs
yihozhang: Good catch! Should be fixed

src/util.rs
```diff
@@ -785,7 +785,7 @@ impl Run {
 
         let executable = self.output_path.clone().unwrap_or_else(|| self.name());
         let opt_level = if optimize_brillvm { "-O3" } else { "-O1" };
-        std::process::Command::new("clang")
+        std::process::Command::new("cc")
```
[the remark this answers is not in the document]
oflatt: I think you need clang, it runs llvm

oflatt (approved):
Thanks!

PR #447, opened 2024-04-11
Keep single scope context at leaves

oflatt:


dag_in_context/src/schema.egg
```diff
@@ -158,20 +158,14 @@
 ; =================================
 
 (datatype Assumption
+  ; Assume nothing
+  (NoContext)
```
kirstenmg, line 162: Why do we need `NoContext` at all?

dag_in_context/src/utility/in_context.rs
kirstenmg, line 1: Do you have any tests that use `InIf`? I do not understand how information about the predicate is propagated, but I guess not propagating `InIf` will become obsolete once `If`s have regions?

dag_in_context/src/utility/context-prop.egg
```diff
@@ -1,49 +1,69 @@
...
-;; we normalize terms to keep contexts in the leaf nodes.
-;; This file contains the rules that move contexts to leaf nodes
-
+;; we normalize terms by pushing loop and func contexts to leaf nodes.
+;; Pushing InIf contexts down to leaves causes exponential blowup, so we avoid this.
+ 
 (ruleset context-prop)
 
+;; A leaf context is either a InLoop or a NoContext.
+;; We maintain the invariant that there is only one such context per region, and it can be pushed down to
```
kirstenmg, line 8: Should we also add rules to check this invariant? Would the commented-out rule in `context_of.egg` check this invariant?

kirstenmg (approved):
Approve; all concerns resolved by if regions!

PR #448, opened 2024-04-11
Substitution with fuel

oflatt:


dag_in_context/src/utility/subst.egg
```diff
@@ -52,58 +65,58 @@
...
          :ruleset subst)
-(rewrite (Subst assum to (Get c1 index))
-         (Get (Subst assum to c1) index)
+(rewrite (Subst fuel assum to (Get c1 index))
+         (Get (Subst (- fuel 1) assum to c1) index)
          :ruleset subst)
-(rewrite (Subst assum to (Alloc id c1 state ty))
-         (Alloc id (Subst assum to c1) (Subst assum to state) ty)
+(rewrite (Subst fuel assum to (Alloc id c1 state ty))
+         (Alloc id (Subst (- fuel 1) assum to c1) (Subst fuel assum to state) ty)
```
kirstenmg, line 79: Should decrease both fuels

dag_in_context/src/utility/subst.egg
kirstenmg, line 1: Shouldn't we stop substituting when the fuel gets to 0? I don't see where that happens in these rules.

kirstenmg (changes requested):

dag_in_context/src/utility/subst.egg
kirstenmg: Related to this, I think there should be a test for a non-infinite fuel, e.g. a fuel of 2

oflatt:
Don't forget to request a review when this is ready

PR #451, opened 2024-04-12
Fix soundness bug- unioning no context with something that has context

oflatt:
I wish that we got an error instead of a lack of saturation. We should think of some sort of check we can do

### DIFF
--- a/dag_in_context/src/interval_analysis.egg
+++ b/dag_in_context/src/interval_analysis.egg
@@ -63,32 +63,36 @@
        (= (IntB x) (lo-bound expr))
        (= (IntB x) (hi-bound expr))
        (HasArgType expr ty)
+       (ContextOf expr ctx)
       )
-      ((union expr (Const (Int x) ty)))
+      ((union expr (InContext ctx (Const (Int x) ty))))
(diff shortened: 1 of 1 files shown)


dag_in_context/src/interval_analysis.egg
```diff
@@ -63,32 +63,36 @@
...
       :ruleset interval-analysis)
 
 (rule (
        (= (BoolB x) (lo-bound expr))
        (= (BoolB x) (hi-bound expr))
        (HasArgType expr ty)
+       (ContextOf expr ctx)
       )
-      ((union expr (Const (Bool x) ty)))
+      ((union expr (InContext ctx (Const (Bool x) ty))))
```
ajpal, line 77: Oh interesting, I didn't realize this was a problem. What if `expr` isn't in a context?

ajpal (approved):

dag_in_context/src/interval_analysis.egg
oflatt: Everything has a context! I'm writing these invariants down in the code and paper soontm

PR #452, opened 2024-04-12
interval_analysis after context_of

ajpal:
Fix import order issue introduced by https://github.com/egraphs-good/eggcc/pull/451


oflatt:
still needed?

PR #453, opened 2024-04-12
Implement passthrough optimziations

Alex-Fischman:
Adds five rules for basic simplifications.


dag_in_context/src/optimizations/passthrough.egg
```diff
@@ -0,0 +1,47 @@
+;; Pass through thetas
+(rule ((= lhs (Get loop i))
```
oflatt, line 2: For non-terminating loops, this is unsound. We should have a test for this
I think you need to check the type of the state edge, and don't do this optimization if it's a state edge.

dag_in_context/src/optimizations/passthrough.egg
```diff
@@ -0,0 +1,47 @@
...
+       (= branch0 (ListExpr-ith branches 0))
+       (= branch1 (ListExpr-ith branches 1))
+       (= (Get branch0 i) (Get (Arg _) j))
+       (= (Get branch1 i) (Get (Arg _) j))
+       (= passed-through (Get inputs j)))
+      ((union lhs passed-through))
+      :ruleset always-run)
+
+;; Pass through switch predicate
+(rule ((= lhs (Get switch i))
```
oflatt, line 22: cool rule

oflatt (changes requested):

dag_in_context/src/optimizations/passthrough.egg
Alex-Fischman: Hmm. I don't need this for my optimization. Since we're in hack mode, should I just comment this one out with a comment?

Alex-Fischman: I went ahead and did this to move faster. Feel free to unresolve if you want me to do something else.

oflatt (approved):
Looks good
Could you make an issue/item in the timeline for the passthrough rules?

PR #455, opened 2024-04-14
Fix get node cost bug

Alex-Fischman:
Fixes an extraction bug where things were wrong when a node's children were ignored.


oflatt (approved):
Great job on this

PR #456, opened 2024-04-14
Implement cost model

Alex-Fischman:
Implements a cost model for all constructors and relations and datatypes using a match statement and a trait.

It is concerning to me that none of the snapshots changed, but maybe it's fine.


oflatt (approved):
looks good to me

PR #458, opened 2024-04-15
Refactor extraction in preparation for linearity preservation

oflatt:


dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -1,10 +1,76 @@
...
+            types: HashMap::new(),
+            costs: FxHashMap::<ClassId, CostSet>::with_capacity_and_hasher(
+                egraph.classes().len(),
+                Default::default(),
+            ),
+            unextractables,
+        }
+    }
+
+    pub(crate) fn add_types(&mut self) {
```
oflatt, line 44: find the types for expressions so we can use them in linearity checks

oflatt:
This is stale now

PR #463, opened 2024-04-17
Function inlining

kirstenmg:
Implementation of function inlining for DAG-in-context.
I had to carefully manipulate the schedule to make substitution go deep enough without continuing forever on infinitely recursive functions. This is bad, and things may get too slow again once other optimizations are in. However, I don't know what alternative there is other than adding phi nodes.

Note that the majority of `function_inlining.egg` is a copy-pasted Subst, renamed to FunctionInliningSubst. The actual inlining rule is at the end, since it uses FunctionInliningSubst.


dag_in_context/src/schedule.egg
```diff
@@ -1,23 +1,29 @@
...
+          ;; make sure to saturate context-helpers, since they ensure the other rules saturate
+          (saturate context-helpers)
+          context
+          context-prop))
+      ;; TODO enable this ruleset again after fixing for regions
+      ;;conditional-invariant-code-motion
+      switch_rewrite
+      loop-simplify
+      ;; Substitution should go for a few iterations but not saturate
+      (repeat 4 function-inlining-subst))
```
oflatt, line 25: 4 is quite few, we should talk about this

tests/snapshots/files__nested_call-optimize.snap
```diff
@@ -3,21 +3,21 @@ source: tests/files.rs
...
   v5: int = mul v2 v0;
   ret v5;
 }
 @main {
-  v1: int = const 1;
-  v3: int = const 0;
-  v6: int = call @inc v3;
-  print v6;
-  v10: int = call @double v1;
-  print v10;
```
oflatt, line 22: neat!

yihozhang (approved):

kirstenmg:
Rewriting this PR to do function inlining in Rust.

yihozhang (approved):
LGTM!

dag_in_context/src/optimizations/function_inlining.rs
```diff
@@ -1 +1,83 @@
...
+
+#[derive(Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
+pub struct CallBody {
+    pub call: RcExpr,
+    pub body: RcExpr,
+}
+
+// Gets a list of all the calls in the program
+// and pairs them with an inlined body
+fn get_calls_and_subst(
```
oflatt, line 13: I think you should separate concerns- make a `get_calls` function and use the resulting set to do the inlining

dag_in_context/src/schema_helpers.rs
```diff
@@ -206,6 +233,166 @@ impl Expr {
...
+                Self::subst_with_cache(arg, arg_ty, arg_ctx, x, subst_cache),
+                Self::subst_with_cache(arg, arg_ty, arg_ctx, y, subst_cache),
+            )),
+            Expr::If(pred, input, then, els) => {
+                let new_pred = Self::subst_with_cache(arg, arg_ty, arg_ctx, pred, subst_cache);
+                let new_input = Self::subst_with_cache(arg, arg_ty, arg_ctx, input, subst_cache);
+                Rc::new(Expr::If(
+                    new_pred.clone(),
+                    new_input.clone(),
+                    then.clone(),
```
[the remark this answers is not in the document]
kirstenmg: Using `replace_ctx` for if/switch branches or loop bodies is much slower. To generate egglog for `fib_recursive.bril`, it's < 1 s without replacing vs. > 1 min with replacing. This is because replacing context creates more calls to inline. Previously-inlined calls in branches and loop bodies are not equivalent to the context-replaced calls, so the context-replaced calls will also have their corresponding bodies substituted and added to the inlining unions.

If it is still correct to not replace the context, I think it is better to not replace the context and make inlining run in a reasonable time at the expense of losing some opportunities for optimization.

kirstenmg:
Function inlining only runs for 2 iterations right now (see #494 ), so some bril programs, like `simple_recursive.bril`, do not have their calls optimized away.

dag_in_context/src/lib.rs
```diff
@@ -95,12 +104,72 @@ pub(crate) fn print_with_intermediate_vars(termdag: &TermDag, term: Term) -> Str
     printed
 }
 
+// Takes in a termdag and a shared cache referencing the termdag, which can be re-used for multiple exprs
+fn global_cached_print_expr_with_intermediate_helper(
```
oflatt, line 108: This has a pretty confusing name- I think you should inline this function so it doesn't exist

oflatt (changes requested):
Really great work! Fix snapshots and I'll merge this

tests/snapshots/files__unstructured-optimize.snap
```diff
@@ -3,7 +3,7 @@ source: tests/files.rs
 expression: visualization.result
 ---
 @main(v0: int) {
-  v3: bool = const true;
+  v3: bool = const false;
```
kirstenmg, line 6: This change seems a bit odd, but it was originally false: https://github.com/egraphs-good/eggcc/blob/86d79ea803034425ca1f81c5167950297083e6d1/tests/snapshots/files__unstructured-optimize.snap

oflatt: 🤔 I've seen that change before, it probably doesn't matter for some reason

PR #464, opened 2024-04-17
[Interval Analysis] Actually working on a bril program!

ajpal:
```
# ARGS: 1
@main(input: int) {
  zero: int = const 0;
  cond: bool = le input zero;
  br cond .then .else;
  .then:
    y: int = id input;
    jmp .end;
  .else:
    neg: int = const -1;
    y: int = mul input neg;
    jmp .end;
  .end:
    z: bool = le y zero;
    print z;
}
```

Becomes

```
@main(v0: int) {
  v2: bool = const true;
  print v2;
}
```


dag_in_context/src/optimizations/passthrough.egg
```diff
@@ -8,6 +8,7 @@
 ;       :ruleset always-run)
 
 ;; Pass through switch arguments
+;; TODO: Similar fix as below with InContext nodes
```
Alex-Fischman, line 11: Can you just do this? It should be a two-line diff, where you just wrap the `Args` in `InContexts`, right? If it's harder for some reason just put that reason in the comment.

dag_in_context/src/interval_analysis.egg
```diff
@@ -249,4 +265,73 @@
...
+        (= expr (Get inputs i))
+        ; the right operand of the < has an upper bound
+        (= (IntB v) (hi-bound value))
+        ; context node inside the if region
+        (= ctx (InContext (InIf true pred inputs) (Arg ty)))
+        (HasType inputs ty)
+        ; usage of the value x inside the if region
+        (Get ctx i)
+      )
+      ((set (hi-bound (Get ctx i)) (IntB (- v 1)))) ; expr < value was true, so we know expr < (hi-bound value) - 1
```
Alex-Fischman, line 287: Why is there a `- 1` here? If value is 2 and we know expr<2, expr could still be 1, right?

dag_in_context/src/interval_analysis.egg
```diff
@@ -249,4 +265,73 @@
...
+        (= expr (Get inputs i))
+        ; the left operand of the < has a lower bound
+        (= (IntB v) (lo-bound value))
+        ; context node inside the if region
+        (= ctx (InContext (InIf true pred inputs) (Arg ty)))
+        (HasType inputs ty)
+        ; usage of the value x inside the if region
+        (Get ctx i)
+      )
+      ((set (lo-bound (Get ctx i)) (IntB (+ v 1)))) ; value < expr was true, so we know expr >= (lo-bound value) + 1
```
Alex-Fischman, line 320: They're equivalent, so why not write expr>lo value instead of expr>=lo value + 1?

dag_in_context/src/interval_analysis.egg
```diff
@@ -249,4 +265,73 @@
...
+        ; the left operand of the < has an upper bound
+        (= (IntB v) (hi-bound value))
+        ; context node inside the if region
+        (= ctx (InContext (InIf false pred inputs) (Arg ty)))
+        (HasType inputs ty)
+        ; usage of the value x inside the if region
+        (Get ctx i)
+      )
+      ((set (hi-bound (Get ctx i)) (IntB v))) ; value < expr was false, so we know expr <= (hi-bound value)
+      :ruleset interval-analysis)
```
Alex-Fischman, line 337: nit: I think that this is a case where metaprogramming would be useful. These four rules all match on approximately the same expression, so it might be valuable to refactor the rule out to Rust. With that said we are in speed mode so no need if you think it'll take too long.

dag_in_context/src/interval_analysis.egg
```diff
@@ -249,4 +265,73 @@
...
+        (= if_e (If pred inputs then else))
+        ; the left operand of the < is an input to the if region
+        (= expr (Get inputs i))
+        ; the right operand of the < has an upper bound
+        (= (IntB v) (hi-bound value))
+        ; context node inside the if region
+        (= ctx (InContext (InIf true pred inputs) (Arg ty)))
+        (HasType inputs ty)
+        ; usage of the value x inside the if region
+        (Get ctx i)
```
oflatt, line 285: I think we don't need to match on this `Get`, the rule will still work

dag_in_context/src/interval_analysis.egg
oflatt: I think it's correct- the hi-bound is inclusive, so this code sets the hi-bound to 1.

oflatt, line 287: One thing I'm concerned about with the current interval analysis is integer overflow
What happens when this overflows? I think it makes the interval analysis unsound
A simple fix would be to add to the query that v is greater than INT_MIN

oflatt: All the other interval rules probably have this problem, so lets not solve it in this PR.
We don't necessarily need to solve it unless we find it happens in benchmarks

dag_in_context/src/interval_analysis.egg
oflatt: I recall you were working on these sort of normalization rules- did they end up having a problem?

dag_in_context/src/interval_analysis.egg
oflatt: I'd rather we normalize the expressions and have just one rule that does this

dag_in_context/src/optimizations/passthrough.egg
```diff
@@ -31,12 +32,15 @@
...
 
 ;; Pass through if arguments
-(rule ((= lhs (Get if i))
+(rule ((= if_get (Get if i))
        (= if (If pred inputs then_ else_))
-       (= (Get else_ i) (Get (Arg _) j))
-       (= (Get then_ i) (Get (Arg _) j))
-       (= passed-through (Get inputs j)))
-      ((union lhs passed-through))
+       (= (Get then_ i) (Get (InContext (InIf true pred inputs) (Arg ty)) j))
```
oflatt, line 37: Along the same lines as alex, I think you can match on any old context in here- doesn't need to be InIf

tests/snapshots/files__if_context-optimize.snap
```diff
@@ -3,19 +3,6 @@ source: tests/files.rs
...
-  jmp .__25__;
-.__11__:
-  v13: int = const -1;
-  v15: int = mul v0 v13;
-  v18: int = id v15;
-  v19: int = id v3;
-.__25__:
-  v27: bool = lt v18 v19;
-  print v27;
+  v2: bool = const true;
```
oflatt, line 6: so cool!

oflatt (changes requested):
Great work! Mostly had questions around the code duplication

dag_in_context/src/interval_analysis.egg
ajpal: I wrote normalization rules that phrase everything in terms of < (not <=, >, >=)

dag_in_context/src/interval_analysis.egg
ajpal: We can't do that- we normalize everything to use `<`, but we still have four cases:
`a < b` where you know a bound of `b`
`a < b` where you know a bound of `a`
and for each of those, you have a rule for the true case and the false case.

dag_in_context/src/interval_analysis.egg
ajpal: Will handle as follow-up PR

dag_in_context/src/optimizations/passthrough.egg
Alex-Fischman: In that case I'll deal with it when I do the theta passthrough optimization.

dag_in_context/src/interval_analysis.egg
Alex-Fischman: Then the code is right but the comment is wrong.

dag_in_context/src/interval_analysis.egg
Alex-Fischman: I get it now, it's because the lo-bound is inclusive. Just change the comment in the previous convo.

Alex-Fischman:
@ajpal I did the passthrough stuff in a separate PR so that I'm not blocked on this one, rebase on main once #466 is in.

oflatt (approved):
Great! Very happy with this one

PR #471, opened 2024-04-19
Improved branch restructuring

ezrosent:
This PR adapts the algorithm from the original RVSDG papers for restructuring branches. This will let us translate more programs successfully.

I suspect there is more work to do on the translation back to CFGs now that this is up.

This PR also includes a few cleanups around making LLVM tests optional, and adding some switches to dump CFG visualizations to `/tmp`; both of those use cargo features.


src/rvsdg/restructure.rs
```diff
@@ -306,93 +276,242 @@ impl SwitchCfgFunction {
...
 }
 
 const JMP: Branch = Branch {
     op: BranchOp::Jmp,
     pos: None,
 };
+
+#[derive(Default)]
+struct Continuation {
+    // Nodes in the "tail" that are targetted by an edge out of the given branch node.
```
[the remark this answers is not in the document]
oflatt: I think the tail is the dominance frontier or something like that?

src/rvsdg/restructure.rs
```diff
@@ -306,93 +276,242 @@ impl SwitchCfgFunction {
...
+        if cont.reentry_nodes.is_empty() {
+            // Nothing to do.
+            return;
+        }
+        // assert!(
+        //     !cont.reentry_nodes.is_empty(),
+        //     "No reentry nodes found [start: {start:?}, entry: {entry:?}, exit: {exit:?}]"
+        // );
+
+        if cont.reentry_nodes.len() == 1 {
```
oflatt, line 380: Which of these if statements are "optimizations" that generate better cfgs and which ones are necessary? For example, is this one necessary or does the else case technically work?

src/rvsdg/restructure.rs
```diff
@@ -306,93 +276,242 @@ impl SwitchCfgFunction {
...
+                    self.split_arc(edge);
+                    continue;
                 }
-            }
 
-            // The rest of this loop is focused on mux/demux for multiple
-            // continuations. If we only have one, we are done.
-            if conts.len() == 1 {
-                continue;
+                if reentry_edges.len() == 1 {
```
oflatt, line 403: is this one an "optimization"?

src/rvsdg/restructure.rs
```diff
@@ -306,93 +276,242 @@ impl SwitchCfgFunction {
...
+                    continue;
+                }
+                // There are multiple exit edges to the tail. Have them all join
+                // into an intermediate block.
+                let inter = self.fresh_block();
+                self.graph.add_edge(inter, tail, JMP);
+                for e in reentry_edges {
+                    self.divert_edge(*e, inter);
+                }
+                self.restructure_branches_inner(target, inter, state);
```
oflatt, line 421: In this recursive case, do we necessarily hit the if statement that came before this?
If so, can that be a helper that we call here?

src/rvsdg/from_cfg.rs
```diff
@@ -331,32 +399,14 @@ impl<'a> RvsdgBuilder<'a> {
...
-                        };
-                        !dom.any(|n| n == curr)
-                    })
-                    .nth(1)
-                    .is_some()
-                {
-                    break;
-                }
+            while curr != join_point {
                 curr = self.try_loop(curr)?.unwrap();
```
oflatt, line 410: What does it mean for try_loop or try_branch to return None?

oflatt (changes requested):
Some questions- I think I'll need to digest this algorithm better sometime, but it should't block the PR.
However, I'd love to see at least a few more tests before we merge. Perhaps more unstructured bril programs that don't involve pointers? In fact, we could randomly generate some cfgs and see what happens

src/rvsdg/restructure.rs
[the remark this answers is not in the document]
ezrosent: Added a comment here.

src/rvsdg/restructure.rs
```diff
@@ -306,93 +276,242 @@ impl SwitchCfgFunction {
...
+    }
 }
 
 const JMP: Branch = Branch {
     op: BranchOp::Jmp,
     pos: None,
 };
+
+#[derive(Default)]
+struct Continuation {
```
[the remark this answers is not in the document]
ezrosent: Done!

src/rvsdg/restructure.rs
```diff
@@ -306,93 +276,242 @@ impl SwitchCfgFunction {
...
+                break;
+            };
+            if neighbors.next().is_some() {
+                break;
+            }
+            cur = next;
+        }
+        cur
+    }
+    fn divert_edge(&mut self, edge: EdgeIndex, new_target: NodeIndex) -> EdgeIndex {
```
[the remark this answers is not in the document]
ezrosent: Done. I also renamed the method to `retarget`.

src/rvsdg/restructure.rs
ezrosent: Yup. Same answer as the other question.

src/rvsdg/restructure.rs
ezrosent: Added some comments here. This structure is also in the JLM repo: it avoids us adding extra nodes and extra predicates when we don't need to.

src/rvsdg/restructure.rs
ezrosent: I don't think so. We don't actually know anything about target here, just that it isn't dominated by any of the outgoing edges. Unless I'm misunderstanding which `if` you're talking about.

src/rvsdg/from_cfg.rs
ezrosent: It happens if you reach the exit block early.

ezrosent:
> However, I'd love to see at least a few more tests before we merge. Perhaps more unstructured bril programs that don't involve pointers? In fact, we could randomly generate some cfgs and see what happens

There are 3 or 4 more failing CFGs in the issues that I'm hoping to fix in follow-ups, but I can look at adding some more.

ezrosent:
I'm not sure where the failing test is from either... that snapshot is present on this branch.

oflatt (approved):
Great work!

PR #476, opened 2024-04-20
Make DAG extractor deterministic

rtjoa:
Previously, the DAG extractor relied on the iteration order of HashMaps.

I tested this with the following (previously, `fib_recursive-optimize` was nondeterministic and occasionally failed snapshot tests).
```
for _ in {1..100}; do
  if ! cargo test -r fib_recursive-optimize; then
    exit 1
  fi
done
```


Alex-Fischman (commented):
LGTM

oflatt (approved):
Good fix- I'll change this to a indexmap sometime for efficiency

PR #479, opened 2024-04-22
Optimize extraction algorithm using worklist, enabling medium-sized benchmarks

oflatt:
This PR optimizes the extraction algorithm to scale to medium-sized benchmarks.
It uses a parent table, similar to extraction algorithms in the extraction-gym

### DIFF
--- a/dag_in_context/src/greedy_dag_extractor.rs
+++ b/dag_in_context/src/greedy_dag_extractor.rs
@@ -10,11 +10,14 @@ use crate::{
     typechecker::TypeChecker,
 };
 
+type RootId = ClassId;
+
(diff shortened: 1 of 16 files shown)


yihozhang (approved):
This looks good! One thing I want to note though is that, looking at the diff, the previous extraction algorithm updates cost on a region-to-region basis, while the new algorithm is more like a bottom-up traversal and does not care about regions. Although this is different than what we previously thought, I don't think this causes any big issue.

yihozhang:
Also, we don't really need to distinguish `root_id` for the first extraction pass. After the whole algorithm is merged we should refactor this so that the first pass does not distinguish between the same nodes under different regions.

PR #480, opened 2024-04-22
Nightly fix for benchmarks

glenn-sun:
Now all passing benchmarks and tests should appear in nightly. Nightly may take significantly longer to run.


oflatt (approved):
nice!
before merge, confirm this works in a nightly.
You can manually run them here: https://nightly.cs.washington.edu/

oflatt:
Looks like the nightly for this branch is quite slow- we'll want to tackle that issue before merging

glenn-sun:
@oflatt Runs in 12 minutes now, with hyperfine running ~100-200 iterations on only the /benchmarks folder. Plenty of buffer within the 1 hr target for when I add more benchmarks. Though the numbers for bril benchmarks are a bit too noisy to be meaningful. I'll click merge if it's okay with you now.

oflatt:
Awesome!

PR #483, opened 2024-04-23
Fix failing tests in `brils/core`

ezrosent:
This PR moves all tests over from the `core` subdir to passing. Some of these were fixed by #471, the rest of the fixes are routine.


oflatt (approved):
Nice fix!

PR #485, opened 2024-04-24
Fix serialized egraph in egglog

oflatt:
See this PR: https://github.com/egraphs-good/egglog/pull/368


oflatt:
This is no longer necessary for now- I fixed this using subsume instead of deletion

PR #490, opened 2024-04-25
Remove `failing/brils/mem`: all tests pass now

ezrosent:
Most of these tests were already passing. One (`fib.bril`) was not present in the passing tests, but it passes at head.


oflatt (approved):
Thanks!

PR #495, opened 2024-04-26
Improve substitution to work when nested using IsResolved

oflatt:


rtjoa (approved):
Happy to see the metaprogramming 🙂. And nice explanatory comment in `is-resolved.rs`.

PR #496, opened 2024-04-26
Three quarters

Alex-Fischman:
Adds rules for the 3/4 if rewrites as well as a few other canonicalization rules.


oflatt (approved):
great!

PR #497, opened 2024-04-27
Fix some minor issues in CFG translation:

ezrosent:
* Don't emit branches if they happen after a `ret`
* Add type information to return variable

This fixes the `relative-primes` benchmark


oflatt (approved):
thanks for the fix!

PR #500, opened 2024-04-28
Test on our benchmarks

oflatt:


src/util.rs
```diff
@@ -376,16 +394,12 @@ impl Run {
         for optimize_egglog in [true, false] {
             for optimize_brilift in [true, false] {
                 for interp in [true, false] {
-                    res.push(Run {
-                        test_type: RunType::CompileBrilift,
+                    Run::compile_brilift_config(
```
Alex-Fischman, line 397: Don't you need to call `res.push` here? I think this loop constructs some `Run`s and then throws them away...

Alex-Fischman (changes requested):

Alex-Fischman (approved):
LGTM

PR #508, opened 2024-04-29
Improve memory usage of extraction, move some benchmarks to passing

oflatt:
This PR also, unfortunately, has to disable the function inlining code which has a performance bug


yihozhang (approved):
Looks good!

dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -354,30 +355,59 @@ fn calculate_cost_set(
 
     let mut shared_total = NotNan::new(0.).unwrap();
     let mut unshared_total = info.cm.get_op_cost(&node.op);
-    let mut costs: HashMap<ClassId, NotNan<f64>> = HashMap::default();
+    let mut costs: HashTrieMap<ClassId, NotNan<f64>> = Default::default();
```
yihozhang, line 358: Do you have a sense of how persistent data structures improve performance? Functional data structures in general have a much larger constant factor.

oflatt: We get more sharing in memory- we didn't get much of a speedup as far as I can tell, but memory usage is halved

oflatt: Actually perhaps we did get a big speedup- looking at the numbers below

PR #512, opened 2024-04-30
Save unoptimized and optimized LLVM when `—output_dir` is provided

rtjoa:
For `--run-mode=compile-bril-llvm`, when `--output-dir=DIR` is provided, save `compile-unopt.ll` and `compile.ll` to `DIR`.

Note that when these options are provided, it runs clang a second time, as I don't think it's possible to dump llvm and the binary in the same command. An alternative would be to break up `llc` from the other stages of `clang`, but that would lead to more complexity in the common case.

Example:
```bash
$ cargo run benchmarks/passing/bril/mem/fib.bril --optimize-egglog=true --optimize-bril-llvm=true --run-mode=compile-bril-llvm --output-dir=output
$ ls output
compile-unopt.ll compile.ll
```


ryan-berger (approved):
LGTM

src/main.rs
```diff
@@ -32,6 +32,9 @@ struct Args {
     /// then the executable will be in `abc`.
     #[clap(short)]
     output_path: Option<String>,
+    /// Where to put intermediary files (only for OptimizeBrilLLVM mode)
+    #[clap(long)]
+    output_dir: Option<String>,
```
Alex-Fischman, line 37: nit: rename this to something less similar to output path, or just reuse output path

Alex-Fischman (commented):
One nit

PR #515, opened 2024-04-30
highlight min and max in each column

ajpal:
image

Kind of a hack in table.js, but I think it's probably fine (esp given deadline etc)


ryan-berger (approved):
LGTM

oflatt (approved):
Looks good to me!

PR #516, opened 2024-04-30
Make diff tool work in nightly again

oflatt:
This PR does three things (sorry)
1. It makes the diff tool work again by adding text to the table
2. It generates warnings for missing benchmarks
3. It adds a `localnightly.sh` script that you can use to run a local nightly on a particular directory

Here's my local output of `bash localnightly.sh benchmarks/passing/bril/core/gcd.bril`
image


oflatt:
Don't review yet! Fixing conflicts

infra/localnightly.sh
```diff
@@ -0,0 +1,3 @@
+# pass arguments to nightly.sh
```
ajpal, line 1: Why does this need to be a separate script? Or, can it be python instead of bash?

infra/nightly-resources/index.js
```diff
@@ -204,6 +243,17 @@ async function loadBenchmarks(compareTo) {
...
     let container = document.getElementById("profile");
     container.innerHTML = ConvertJsonToTable(parsed);
+
+    // Add warnings
+    let warningContainer = document.getElementById("warnings");
+    warningContainer.innerHTML = "";
+    warnings.forEach((warning) => {
+        // make warnings red
+        let warningElement = document.createElement("p");
+        warningElement.style.color = "red";
```
ajpal, line 253: Can you instead just do this with css on the container?

infra/nightly-resources/index.js
```diff
@@ -126,34 +132,50 @@ function groupByBenchmark(benchList) {
...
 }
 
+// b may be undefined
+function compareTo(a, b) {
+    // if b is undefined, return a
+    if (b === undefined) {
+        return tryRound(a);
+    } else {
+        var difference = a - b;
+        var sign = difference < 0 ? "" : "+";
```
ajpal, line 142: why no `-` if it improves over main?

infra/nightly-resources/index.js
```diff
@@ -126,34 +132,50 @@ function groupByBenchmark(benchList) {
     return groupedBy;
 }
 
+// b may be undefined
+function compareTo(a, b) {
```
ajpal, line 136: rename: `getDifference` ?

ajpal: compareTo implies boolean IMO

infra/nightly-resources/index.js
```diff
@@ -126,34 +132,50 @@ function groupByBenchmark(benchList) {
...
+        var difference = a - b;
+        var sign = difference < 0 ? "" : "+";
+        // put the difference in parens after a
+        return tryRound(a) + " (" + sign + tryRound(difference) + ")";
+    }
+}
+
+function compareAttribute(results, prevResults, attribute) {
+    const a = results[attribute];
+    const b = prevResults ? prevResults[attribute] : undefined;
```
ajpal, line 150: nit: cleaner to use optional chaining:
`const b = prevResults?.[attribute]`

infra/nightly-resources/index.js
```diff
@@ -126,34 +132,50 @@ function groupByBenchmark(benchList) {
...
+    return { class: "", value: compareTo(a, b)};
+}
+
 const compareKeys = ["# Instructions"];
-function buildEntry(run) {
+// prevRun may be undefined
+function buildEntry(prevRun, run) {
     const results = run.hyperfine.results[0];
+    const prevResults = prevRun ? prevRun.hyperfine.results[0] : undefined;
```
ajpal, line 158: same comment about optional chaining

infra/profile.py
```diff
@@ -1,51 +1,62 @@
...
-    
-  for mode in modes:
-    (name, runmode, options) = mode
-    subprocess.call(f'cargo run --release {profile} --run-mode {runmode} {options} -o {profile_dir}/{name}', shell=True)
-
-    with open(f'{profile_dir}/{name}-args') as f:
-      args = f.read().rstrip()
-    
-    subprocess.call(f'hyperfine --warmup 2 --max-runs 100 --export-json {profile_dir}/{name}.json "{profile_dir}/{name} {args}"', shell=True)
+    # strip the path to just the file name
```
ajpal, line 38: Are there changes here or just whitespace?

ryan-berger: It looks like Oliver's editor turned our two spaces into 4 :(

ajpal: @oflatt It would be good to avoid thrashing the diff with whitespace changes. Can you please revert auto-formatting changes? If you want, we should enforce formatting rules with a linter or something, but until it's enforced, we shouldn't commit changes like this since it makes the history harder to look through.

oflatt:
new screenshot
image

infra/nightly-resources/index.js
```diff
@@ -132,47 +132,59 @@ function groupByBenchmark(benchList) {
...
     return {
         name: run.runMethod,
-        mean: compareAttribute(results, prevResults, "mean"),
-        min: compareAttribute(results, prevResults, "min"),
-        max: compareAttribute(results, prevResults, "max"),
-        median: compareAttribute(results, prevResults, "median"),
-        stddev: compareAttribute(results, prevResults, "stddev"),
-        
+        mean: { class: "", value: tryRound(results.mean) },
+        meanVsBaseline: diffAttribute(results, baselineResults, "mean"),
```
ajpal, line 172: nit: name `vsMain` to be more compact

infra/nightly-resources/index.js
```diff
@@ -126,34 +132,50 @@ function groupByBenchmark(benchList) {
...
+// b may be undefined
+function compareTo(a, b) {
+    // if b is undefined, return a
+    if (b === undefined) {
+        return tryRound(a);
+    } else {
+        var difference = a - b;
+        var sign = difference < 0 ? "" : "+";
+        // put the difference in parens after a
+        return tryRound(a) + " (" + sign + tryRound(difference) + ")";
```
ryan-berger: Not sure how I feel about this implicit type conversion. Can we use a format string?

infra/localnightly.sh
oflatt: We would bounce from python to bash and back to python again, since the nightly script is in bash
Lets keep this in mind if this script gets any more complex

infra/nightly-resources/index.js
oflatt: added comment- - is there for negative number printing already

infra/nightly-resources/index.js
oflatt: this is a dictionary right? There would be duplicate keys

infra/profile.py
oflatt: yes, sorry I didn't realize I had formatting enabled.

infra/nightly-resources/index.js
```diff
@@ -126,34 +132,62 @@ function groupByBenchmark(benchList) {
...
+        return { class: "max", value: "N/A" };
+    } else {
+        var difference = current - baseline;
+        // if the difference is negative it will already have a "-"
+        var sign = difference < 0 ? "" : "+";
+        var color = "";
+        if (difference < 0) {
+            color = "min";
+        } else if (difference > 0) {
+            color = "max";
```
ajpal, line 149: nit: should change these class names to `good` / `bad` or something now that they're not always just for mins/maxes

infra/nightly-resources/index.js
```diff
@@ -126,34 +132,62 @@ function groupByBenchmark(benchList) {
...
+// If baseline_number is undefined, it will return N/A
+function getDifference(current, baseline) {
+    // if b is undefined, return a
+    if (baseline === undefined) {
+        return { class: "max", value: "N/A" };
+    } else {
+        var difference = current - baseline;
+        // if the difference is negative it will already have a "-"
+        var sign = difference < 0 ? "" : "+";
+        var color = "";
```
ajpal, line 145: nit: name this variable `cssClass` or something, it doesn't contain a color.

infra/nightly-resources/index.js
```diff
@@ -165,10 +187,17 @@ async function loadBenchmarks(compareTo) {
...
                 return undefined;
             }
+            // prevRun may be undefined
             const prevBenchmark = previousRun ? previousRun[benchName] : undefined;
-            const prevRun = prevBenchmark ? prevBenchmark[runMethod] : undefined;
+            if (prevBenchmark === undefined) {
+                addWarning(`Couldn't find a previous benchmark for ${benchName}`);
+            }   
+            const prevRunForMethod = prevBenchmark ? prevBenchmark[runMethod] : undefined;
+            if (prevBenchmark && prevRunForMethod === undefined) {
```
ryan-berger: why `=== undefined` in the previous if and just `prevBenchmark` here? I'm cool with exploiting the fact that undefined/null is false-y, but imo it should be standard. Either be explicit about `=== undefined` or not, but not both

infra/nightly-resources/index.js
```diff
@@ -258,11 +258,12 @@ async function loadBenchmarks(compareTo) {
 
     // Add warnings
     let warningContainer = document.getElementById("warnings");
+    // make warning red
+    warningContainer.style.color = "red";
```
ajpal, line 262: Sorry, that's not what I mean. Put a class on the html element in index.html, and add a rule in stylesheet.css. Doing style inline like this is a code smell we should avoid when possible.

infra/localnightly.sh
ryan-berger: Alternatively, why can't this be in the make file like it was previously? I'm not sure I understand wanting to throw this small thing into a bash script.

I think a `make serve-nightly` might be better to run only the server, but am open to whatever

ryan-berger (commented):
Few comments

infra/nightly-resources/index.js
ajpal: +1

infra/localnightly.sh
oflatt: It takes an argument, which make doesn't support ):

ryan-berger: Ahhhhhh, I think I see why it is an issue. Is there a reason it can't be an env var instead of an argument?

oflatt: reason being I really don't like env vars

ryan-berger: I personally like the idea of a default variable inside of the make file that'll run the local nightlies with everything, but then you can overwrite it with an env var

infra/nightly-resources/index.js
```diff
@@ -258,12 +258,12 @@ async function loadBenchmarks(compareTo) {
...
     let warningContainer = document.getElementById("warnings");
-    // make warning red
-    warningContainer.style.color = "red";
+    
     warningContainer.innerHTML = "";
     warnings.forEach((warning) => {
-        // make warnings red
         let warningElement = document.createElement("p");
+        // make warning red with warning class
+        warningElement.classList.add("warning");
```
ajpal, line 266: Instead of this, can't you just add this in the css? I don't like manipulating things like this in the JS, it's better to just put the right selectors and stuff in the html file.
```
#warnings {
  color: red;
}
```

infra/nightly-resources/index.js
oflatt: good idea

ajpal (approved):

PR #518, opened 2024-04-30
Loop unrolling working on egglog test

oflatt:


tests/snapshots/files__loop_pass_through-optimize.snap
```diff
@@ -3,7 +3,7 @@ source: tests/files.rs
 expression: visualization.result
 ---
 @main(v0: int) {
-  v1_: int = const 1;
+  v1_: int = const 2;
```
ezrosent, line 6: Why did this value change? It looks like we're comparing the loop variable against 10, but the rule looks for multiples of 4?

dag_in_context/src/optimizations/loop_unroll.egg
```diff
@@ -23,3 +23,44 @@
           outputs) ;; right now, loop unrolling shares the same outputs, but we could add more context here
         (Arg inputs-ty (NoContext)))))
  :ruleset loop-peel)
+
+
+(function UnrollCtx (Expr Expr) Assumption)
+
+;; unroll a loop with constant bounds and initial value
+(rule
```
ezrosent, line 31: This rule looks reasonable. Question though:

I was imagining that loop unrolling would do something like:

```
do {
   <body>
} while (<pred>)
=>
do {
  <body>
  if (<pred>) <body>
} while(<pred>)
```

I think we'd be able to rewrite bounded-iteration loops this way, and we'd match more often. If we really are comparing (i < C) and (i + 1 < C) ... we'd be able to simplify predicates as well I'd imagine.

Why are we doing this one that hard-codes a multiple of 4? I suspect there's a subtlety I'm missing.

ezrosent (approved):
couple clarification questions.

dag_in_context/src/optimizations/loop_unroll.egg
oflatt: I'd like to make it work that way instead- I decided to do the simplest thing first

tests/snapshots/files__loop_pass_through-optimize.snap
oflatt: Good catch, I'm not sure... I can't reproduce it locally

oflatt: I see now- it's equivalent because you can loop peel the loop once.

ezrosent: But this rule shouldn't do peeling, right? It shouldn't match.

But if you can't repro locally maybe it was a different rule that caused it?

PR #520, opened 2024-04-30
Small improvements to extraction

yihozhang:
* Add a test case where first extraction failed but second succeed
* Don't do second extraction if the first one is successful


oflatt (approved):
Great!

PR #521, opened 2024-04-30
Remove list of helpers from extraction

oflatt:
If we change the schema, it'll be fine because we'll get an error that something wasn't extractable.
I also optimized the extraction algorithm to not consider parents that have infinite cost


yihozhang:
This feels like the same logic as the very first version (namely non-existence means 0/infinity). If the user adds a new operator (say `fma`) without an explicit cost, it's still possible to extract an alternative program that doesn't use `fma` and fail silently.

oflatt:
I've strengthened our tests to catch this case.
When someone adds a new operator (and a test for that operator), we'll fail to extract it because the program will have infinite cost.

Alex-Fischman (approved):
LGTM

PR #522, opened 2024-05-01
Remove dependency on NoContext in loop unrolling

yihozhang:
This is the only place where NoContext is used in optimizations. Since we are converting from NoContext to InFunc as the "base" context, we need to make sure our optimizations do not rely on it in critical ways.


oflatt (approved):
Looks good

PR #523, opened 2024-05-01
Add some basic peepholes

ezrosent:
This change starts a basic peephole optimization files. Most moderately complex peepholes I've found rely on things like shifting operators, sign extensions, etc. None of those can be done at the bril level (but note that LLVM will take advantage of them with optimizations turned on!)

Still, these probably don't hurt. I've added in some rules for forwarding predicates for ifs. Let me know if I'm using contexts correctly here.

### DIFF
--- a/dag_in_context/src/lib.rs
+++ b/dag_in_context/src/lib.rs
@@ -52,6 +52,7 @@ pub fn prologue() -> String {
         include_str!("utility/expr_size.egg"),
         include_str!("interval_analysis.egg"),
         include_str!("optimizations/switch_rewrites.egg"),
+        include_str!("optimizations/peepholes.egg"),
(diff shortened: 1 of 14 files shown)


oflatt (changes requested):
One concern- I noticed you added the ruleset to the "saturating" ruleset.
Are we completely certain these always saturate? I think it would be safer to add them to the "optimizations" ruleset, or a new "pre-optimizations" ruleset

ezrosent:
Done! I think these ones do saturate but in general they may not. Also optimizations is a better description of what the goal is.

ezrosent:
Looks like there's a failing test: `loop_invariant::test_invariant_detect` 

I'll take a look

oflatt (approved):
Yay!

PR #524, opened 2024-05-01
Nightly cleanup

ajpal:
Functional changes:
1. Make diff + checkboxes interact correctly (this was broken by https://github.com/egraphs-good/eggcc/pull/516)
2. Hide warnings by default

Code structure changes:
1. move event handling logic into handlers.js
2. Separate data fetching logic from rendering logic
3. Autoformat everything and enforce formatting in the Makefile
4. Be more clear about what the global/persistent data structure is

### DIFF
--- a/infra/nightly-resources/index.js
+++ b/infra/nightly-resources/index.js
@@ -1,25 +1,27 @@
 
 
 // copied from profile.py
-const allmodes = [["rvsdg_roundtrip", "rvsdg-round-trip-to-executable", ""],
+const allModes = [
+    "rvsdg_roundtrip",
+    "egglog_noopt_brilift_noopt",
+    "egglog_noopt_brilift_opt",
+    "egglog_opt_brilift_noopt",
+    "egglog_opt_brilift_opt",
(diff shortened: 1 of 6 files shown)


infra/nightly-resources/index.js
```diff
@@ -1,346 +1,313 @@
...
+];
+
+const GLOBAL_DATA = {
+  enabledModes: new Set(),
+  warnings: new Set(),
+  currentRun: {},
+  baselineRun: undefined,
+};
 
 function addWarning(warning) {
```
oflatt, line 21: should addWarning be in handlers? not sure the precise destinction

oflatt (approved):
Thank you! This looks much better
Perhaps we should have a single command that formats everything, including rust code? It could run `cargo fmt` in both directories and prettier on `js` and whatever on python

infra/nightly-resources/index.js
ajpal: Nope- handlers is for stuff that gets called from a browser event (like on page load, on dropdown change, checkbox, etc)

PR #525, opened 2024-05-01
Add chart to nightly

ajpal:
image

Also makes benchmarks toggleable:
image
So that you can look at the graph for one benchmark at a time, which is helpful for scaling reasons:
image

### DIFF
--- a/infra/nightly-resources/Plugin.Errorbars.js
+++ b/infra/nightly-resources/Plugin.Errorbars.js
@@ -1,3 +1,5 @@
+// Copied from https://github.com/HandsOnDataViz/chartjs-error-bars
+
 (function (global, factory) {
   typeof exports === "object" && typeof module !== "undefined"
     ? (module.exports = factory(require("chart.js")))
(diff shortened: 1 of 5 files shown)


infra/nightly-resources/Plugin.Errorbars.js
```diff
@@ -0,0 +1,310 @@
+(function (global, factory) {
```
ajpal, line 1: Doesn't need review

infra/nightly-resources/chart.js
```diff
@@ -0,0 +1,72 @@
...
+  egglog_opt_bril_llvm_opt: "brown",
+};
+
+function parseDataForChart() {
+  const benchmarks = Array.from(GLOBAL_DATA.enabledBenchmarks);
+  const datasets = {};
+  GLOBAL_DATA.enabledModes.forEach((mode) => {
+    datasets[mode] = {
+      label: mode,
+      backgroundColor: COLORS[mode],
```
oflatt, line 19: if this index fails do we get an error?

oflatt (approved):
Looks great!
We should make sure we get errors if we change the fields of the data

PR #526, opened 2024-05-01
Simple Loop Strength Reduction

kirstenmg:
Simple loop strength reduction with multiplication by a constant works. Multiplication by an invariant is left for the future due to other priorities. See how the multiplication goes away in `strong_loop.bril`!


dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -637,7 +637,10 @@ impl CostModel for DefaultCostModel {
     }
 
     fn ignore_children(&self, op: &str) -> bool {
-        matches!(op, "InLoop" | "NoContext" | "InSwitch" | "InIf")
+        matches!(
+            op,
+            "InLoop" | "NoContext" | "InSwitch" | "InIf" | "InExtendedLoop"
```
oflatt, line 642: Do you need this? I think probably not

oflatt (approved):
Looks good!

PR #527, opened 2024-05-01
Loop unrolling with egglog-guided extraction

oflatt:


dag_in_context/src/interval_analysis.egg
```diff
@@ -359,3 +359,30 @@
       )
       ((set (hi-bound (Get ctx i)) (IntB hi)))
       :ruleset interval-analysis)
+
+
+(rule (
+       ;; argument has loop context
+       (Arg ty (InLoop inputs outputs))
+       ;; in the loop, the argument is passed through
+       (= (Get (Arg ty some_ctx) ith) (Get outputs (+ 1 ith)))
```
yihozhang, line 368: `some_ctx` must be `(InLoop inputs outputs)`, and this expression is exactly the one you are setting lower bound of, right? We should enforce this (also makes the code look cleaner)

dag_in_context/src/interval_analysis.egg
```diff
@@ -359,3 +359,30 @@
...
+       ;; input has some bound
+       (= bound (hi-bound (Get inputs ith)))
+      )
+      (
+       (set (hi-bound (Get (Arg ty (InLoop inputs outputs)) ith)) bound)
```
yihozhang, line 385: same here

yihozhang (commented):
This looks good overall. I left some comments. Is your code unit-tested, and what snapshot test captures the behavior of loop unrolling/peeling optimization here?

tests/snapshots/files__unroll_multiple_4-optimize.snap
```diff
@@ -3,18 +3,22 @@ source: tests/files.rs
...
-  v6_: int = add v2_ v4_;
-  v7_: bool = lt v6_ v3_;
-  v2_: int = id v6_;
-  v3_: int = id v3_;
+  v2_: int = const 1;
+  v3_: int = id v0_;
+  v4_: int = id v1_;
+  v5_: int = id v2_;
+.v6_:
+  v7_: int = add v3_ v5_;
```
oflatt, line 13: This snapshot captures the behavior- this loop is unrolled

dag_in_context/src/interval_analysis.egg
oflatt: Actually, `some_ctx` is probably a different context since this argument was a part of the context's inputs

dag_in_context/src/optimizations/passthrough.egg
```diff
@@ -1,3 +1,16 @@
+;; Pass through thetas
+(rule ((= lhs (Get loop i))
+        (= loop (DoWhile inputs pred-outputs))
+        (= (Get pred-outputs (+ i 1)) (Get (Arg _ty _ctx) i))
+        (= passed-through (Get inputs i))
```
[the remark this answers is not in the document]
oflatt: good idea

oflatt:
There is no unit test, but the snapshot test is good

yihozhang (approved):

PR #528, opened 2024-05-01
Replace NoContext with InFunc

yihozhang:
so that there will be no sharing among expressions in different functions. A follow-up task is to re-enable function inlining based on this refactor.


oflatt (approved):
Looks good!

PR #530, opened 2024-05-02
Nightly Tex Table

ryan-berger:
This PR generates a tex table similar to what nightlies have on our site, except latex. It also pushes the line count table generator into the nightly infra.

Currently the blocker on this PR would be installing `tokei` which is a line counting too. Should be easily installed with `cargo install tokei`, but I'm unsure how best to get it installed (run it across the system? run it in our script?)


oflatt:
If `cargo install tokei` in the nightly script just works that sounds good to me

infra/nightly.sh
```diff
@@ -51,6 +51,9 @@ mkdir -p ./tmp/bench
 
 rm -r ./tmp/
 
+./infra/gen_linecount.py > "$NIGHTLY_DIR/data/linecount.tex"
```
oflatt, line 54: Don't edit `nightly.sh`
Instead, add to our existing nightly python script.

oflatt (changes requested):
Looks pretty good, but you should add a button to the nightly to download the tex

oflatt (approved):
All looks good! Don't merge until successful nightly

PR #532, opened 2024-05-02
Improve extraction's memory usage by filtering to correct types

oflatt:
The main change in this PR is to check the sort of nodes during the reachability query.
Other than that, this PR refactors costsets to get more sharing.

This PR reduces memory usage by 10x on hard benchmarks


yihozhang (approved):
LGTM!

PR #534, opened 2024-05-03
Fix function inlining performance bug

kirstenmg:
Marking as draft until extraction bug is fixed.


oflatt:
Closing since it was merged already

PR #535, opened 2024-05-06
Fix several extraction bugs

oflatt:


src/util.rs
```diff
@@ -411,6 +396,8 @@ impl Run {
                     ..default
                 };
                 res.push(interp);
+            } else {
```
oflatt, line 399: here we had a terrible bug where some tests were not being run 💀

dag_in_context/src/linearity.rs
```diff
@@ -204,6 +235,8 @@ impl<'a> Extractor<'a> {
                     })
                     .collect();
 
+                let mut effectful_parent: HashMap<*const Expr, RcExpr> = Default::default();
```
oflatt, line 238: added just for debugging

PR #536, opened 2024-05-06
Fix several extraction bugs, enable function inlining

oflatt:
This PR fixes several bugs in extraction, and also includes some fixes to function inlining from @kirstenmg, which allowed us to discover these bugs.

I also had to disable load eliminations because it violated the weakly linear invariant.

### DIFF
--- a/dag_in_context/src/greedy_dag_extractor.rs
+++ b/dag_in_context/src/greedy_dag_extractor.rs
@@ -546,21 +546,22 @@ pub fn extract_with_paths(
         let sort_of_node = info.get_sort_of_eclass(classid);
         // if node is effectful, we only consider it if it is in the effectful path
         if sort_of_node == "Expr" && effectful_paths.is_some() {
-            let is_effectful = extractor.is_eclass_effectful(classid.clone());
-            if is_effectful.is_none() && node.op != "Function" {
+            let effectful_lookup = extractor.is_eclass_effectful(classid.clone());
+            if effectful_lookup.is_none() && node.op != "Function" {
                 // skip when type is unknown
                 continue;
             }
+            if let Some(true) = effectful_lookup {
+                let effectful_nodes = effectful_paths.unwrap().get(&rootid);
+                if effectful_nodes.is_none() {
+                    // continue when this root isn't in effectful_paths
+                    continue;
+                }
 
-            let effectful_nodes = effectful_paths.unwrap().get(&rootid);
-            if effectful_nodes.is_none() {
-                // continue when this root isn't in effectful_paths
-                continue;
-            }
-
-            // skip nodes not on the path
-            if is_effectful.unwrap() && !effectful_nodes.unwrap().contains(&nodeid) {
-                continue;
+                // skip nodes not on the path
+                if !effectful_nodes.unwrap().contains(&nodeid) {
+                    continue;
+                }
             }
         }
 
--- a/dag_in_context/src/greedy_dag_extractor.rs
+++ b/dag_in_context/src/greedy_dag_extractor.rs
@@ -563,7 +563,6 @@ pub fn extract_with_paths(
             } else {
                 // if the op is "Function" we don't need a type
                 if node.op != "Function" {
-                    eprintln!("Failed to get type of node {:?}", node);
                     // skip when type is unknown
                     continue;
                 }
--- a/dag_in_context/src/greedy_dag_extractor.rs
+++ b/dag_in_context/src/greedy_dag_extractor.rs
@@ -555,7 +555,7 @@ pub fn extract_with_paths(
                             continue;
                         }
                     } else {
-                        eprintln!("Root {:?} not found in effectful_paths", rootid);
+                        //eprintln!("Root {:?} not found in effectful_paths", rootid);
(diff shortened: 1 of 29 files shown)


dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -40,11 +40,11 @@ pub(crate) struct Extractor<'a> {
 
     // Each term must correspond to a node in the egraph. We store that here
     // Use an indexmap for deterministic order of iteration
-    pub(crate) correspondence: IndexMap<Term, NodeId>,
+    pub(crate) correspondence: IndexMap<Term, HashSet<NodeId>>,
```
yihozhang, line 43: Isn't there a 1-1 correspondence between term and node_id?

dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -489,18 +497,17 @@ pub fn extract(
...
-            extractor_not_linear.find_effectful_nodes_in_program(&res, &egraph_info);
-        extractor_not_linear.costs.clear();
-        (cost_res, res) = extract_with_paths(
-            extractor_not_linear,
-            &egraph_info,
-            Some(&effectful_nodes_along_path),
-        );
-        extractor_not_linear.check_program_is_linear(&res).unwrap();
-    }
+    let (_cost_res, res) = extract_with_paths(extractor_not_linear, &egraph_info, None);
```
yihozhang, line 500: We don't have to undo this optimization, right?

dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -537,13 +544,26 @@ pub fn extract_with_paths(
...
-            let effectful_nodes = effectful_paths.as_ref().unwrap().get(&rootid).unwrap();
-            if let Some(is_stateful) = extractor.is_node_effectful(nodeid.clone()) {
-                if is_stateful && !effectful_nodes.contains(&nodeid) {
+        // if node is effectful, we only consider it if it is in the effectful path
+        if sort_of_node == "Expr" && effectful_paths.is_some() {
+            if let Some(is_stateful) = extractor.is_eclass_effectful(classid.clone()) {
+                if is_stateful {
+                    if let Some(effectful_nodes) = effectful_paths.as_ref().unwrap().get(&rootid) {
+                        if is_stateful && !effectful_nodes.contains(&nodeid) {
+                            //eprintln!("Node {:?} not found in effectful_paths", node);
```
yihozhang, line 553: remove comments or make it a log::info!

dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -537,13 +544,26 @@ pub fn extract_with_paths(
...
         let sort_of_node = info.get_sort_of_eclass(classid);
-        if sort_of_node == "Expr"
-            && effectful_paths.is_some()
-            && effectful_paths.unwrap().contains_key(&rootid)
-        {
-            let effectful_nodes = effectful_paths.as_ref().unwrap().get(&rootid).unwrap();
-            if let Some(is_stateful) = extractor.is_node_effectful(nodeid.clone()) {
-                if is_stateful && !effectful_nodes.contains(&nodeid) {
+        // if node is effectful, we only consider it if it is in the effectful path
+        if sort_of_node == "Expr" && effectful_paths.is_some() {
```
yihozhang, line 548: This is now getting really complex with 5 (!!) nested ifs. We need to work on this together and simplify the code. Also there are two `is_stateful` in the ifs and one is redundant.

dag_in_context/src/linearity.rs
```diff
@@ -46,59 +46,94 @@ impl<'a> Extractor<'a> {
...
         }
 
-        let mut effectful_classes: HashMap<ClassId, HashSet<NodeId>> = Default::default();
-        for (rootid, expr) in linearity.effectful_nodes {
-            let term = linearity.expr_to_term.get(&expr).unwrap();
-            effectful_classes
-                .entry(rootid.clone())
-                .or_default()
-                .insert(self.node_of(term));
+        let effectful_nodes: HashMap<ClassId, HashSet<NodeId>> = linearity
```
yihozhang, line 61: I'm confused by this diff; let's talk about this

dag_in_context/src/greedy_dag_extractor.rs
oflatt: Nope- since we erase contexts, terms may correspond to multiple nodes

dag_in_context/src/greedy_dag_extractor.rs
oflatt: This optimization was hiding bugs- lets not have it for now

dag_in_context/src/greedy_dag_extractor.rs
oflatt: I agree

oflatt: refactored

yihozhang (approved):
After a discussion with @oflatt , we decide that we will merge this PR. However, this PR currently may create an effectful_path set that is not consistent with the actual extraction result. Will fix this later

PR #537, opened 2024-05-06
Fix bug in loop strength reduction

oflatt:


kirstenmg (changes requested):
Could you add a rule that checks that we never match on `TmpCtx`?
```
(rule ((TmpCtx)) ((panic "TmpCtx should not exist outside rule body")))
```

oflatt:
Great idea, added

kirstenmg (approved):

PR #539, opened 2024-05-06
Stronger testing for benchmarks

oflatt:


benchmarks/passing/bril/core/bitshift.bril
```diff
@@ -1,4 +1,3 @@
-# ARGS: 3 5 10000 4
```
oflatt, line 1: found a bug here

ryan-berger (approved):
LGTM

PR #541, opened 2024-05-07
make nightly fast

ryan-berger:


ryan-berger:
@oflatt Looks like this probably should get closed? Git conflict shows that you've changed the max runs to a smaller number

ryan-berger:
I'll close for now and can reopen if we want it, but I'm not sure this does anything substantive

PR #543, opened 2024-05-08
Peepholes in nonopt llvm

ezrosent:
This is to get us closer to an apples-to-apples comparison with LLVM, given that
most of its peepholes can't be written as rewrites on pure bril.


oflatt:
Great work on this! I'm going to put it on hold while we sort out some bugs in Brillvm. I think we are switching to the rust version of Brilvm, potentially causing conflicts

oflatt (approved):
Lets merge this first: looks like the move to rust brilvm may take longer

PR #544, opened 2024-05-08
Fix multiple terms per eclass bug in extraction

oflatt:
Finding multiple terms for the same eclass results in linearity bugs. This PR fixes that.

### DIFF
--- a/dag_in_context/src/greedy_dag_extractor.rs
+++ b/dag_in_context/src/greedy_dag_extractor.rs
@@ -33,7 +33,6 @@ pub(crate) struct Extractor<'a> {
     pub(crate) termdag: &'a mut TermDag,
     costsets: Vec<CostSet>,
     costsetmemo: FxHashMap<(NodeId, Vec<CostSetIndex>), CostSetIndex>,
-    dummy_contexts: FxHashMap<ClassId, CostSetIndex>,
     costs: FxHashMap<ClassId, FxHashMap<ClassId, CostSetIndex>>,
 
     // use to get the type of an expression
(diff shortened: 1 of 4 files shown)


dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -329,14 +344,72 @@ impl<'a> Extractor<'a> {
...
+        term: Term,
+        node_id: NodeId,
+    ) {
+        if let Some(existing) = correspondence.insert(term.clone(), node_id.clone()) {
+            assert_eq!(existing, node_id);
+        }
+    }
+
+    /// Add `term` to `current_costs`, returning
+    /// 1) a new term that takes advantage of sharing
```
yihozhang, line 363: An example in the comment for the first bullet point can be helpful for understanding, e.g., for E-node `Add(c1, Op(c1))`, The two child cost sets of this E-node can have different witnesses for `c1` and this tries to fix it up by having them agreeing on witnesses.

yihozhang (changes requested):
Overall this looks good to me. The implementation is getting more complex so I think in general we should document extraction-related functions more. I'm also wondering if we can take a top-down approach to extraction [like this](https://github.com/yihozhang/extraction-gym/blob/dppl-sampling-extraction/src/extract/sampling_dag.rs), which I believe won't have the issue to start with (since extractor knows all the sibling cost sets). Happy to chat more if you are interested in refactoring.

dag_in_context/src/greedy_dag_extractor.rs
oflatt: added example

dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -225,30 +223,45 @@ impl<'a> Extractor<'a> {
...
+        // get any node in the class
+        let node_id = info.egraph.classes()[&class_id].nodes.first().unwrap();
+        if let Some(existing) = self.costsetmemo.get(&(node_id.clone(), vec![])) {
+            *existing
+        } else {
+            let dummy = self
+                .termdag
+                .lit(Literal::String(format!("dummy{}", node_id).into()));
+            Self::add_correspondence(&mut self.correspondence, dummy.clone(), node_id.clone());
+            let term = self.termdag.app("InFunc".into(), vec![dummy]);
```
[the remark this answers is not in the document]
oflatt: yeah, added more docs

yihozhang (approved):

PR #548, opened 2024-05-08
Make nightly run egglog in parallel (still benchmarking results sequentially)

oflatt:
My main concern with this PR is that it may not be easily killed by nightly any more. Locally, when I try to use `ctrl+c` it won't die until I spam it ):

Ideas on how to fix this problem welcome

### DIFF
--- a/infra/nightly-resources/index.js
+++ b/infra/nightly-resources/index.js
@@ -293,7 +293,7 @@ function renderWarnings() {
 
   const warningContainer = document.getElementById("warnings");
   warningContainer.innerHTML = "";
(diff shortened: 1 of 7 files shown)


infra/nightly-resources/index.js
```diff
@@ -132,6 +133,12 @@ function findBenchToCompareIdx(benchRuns) {
   throw new Error("Couldn't find a benchmark run from main for comparison");
 }
 
+async function getErrors() {
```
ajpal, line 136: Unused?

ajpal (approved):

PR #549, opened 2024-05-08
Check $EGGCC_ROOT for repo root instead of always assuming cwd

rtjoa:


oflatt (changes requested):
Is there a way to do it relative to the binary? That would be better, we wouldn't need a new env variable

rtjoa:
@oflatt I'm not aware of a general way to do this, and we want to be relative to the source dir, rather than the binary. I think approaches like this are common when there are version-controlled util binaries.

oflatt (approved):

PR #546, opened 2024-05-08
Rebase Brillvm on to main

oflatt:
This PR replaces #542

PR #550, opened 2024-05-08
More informative file names for bril llvm IR dumps

rtjoa:
Instead of `compile-unopt.ll` and `compile.ll`, produce names like `ackermann.ll` and `ackermann_egglog_opt_opt_bril_llvm_O3.ll`.
- Always include the original file name
- For the optimized file:
  - Include `_egglog_opt` iff we optimize with egglog
  - Include clang optimization level


oflatt:
This one is going to have big conflicts with #546 
Can we hold off? Lots of churn

PR #552, opened 2024-05-08
State edge passthrough

Alex-Fischman:
This PR is finally able to fully delete if statements by rewriting them to if statements that don't have state edges if the state edge is passed through. It does this by creating a new subst-like helper called DropAt which can delete an input from a region.


dag_in_context/src/utility/canonicalize.egg
```diff
@@ -67,3 +67,21 @@
          (Concat (Single (Get expr a)) (SubTuple expr (+ a 1) (- b 1)))
          :when ((> b 1))
          :ruleset always-run)
+
+; Helper functions to remove one element from a tuple or type list
+;                           tuple    idx
+(function TupleMinusOne    (Expr     i64) Expr     :unextractable)
```
oflatt, line 73: How about "TupleRemoveAt"?

oflatt (approved):
Looks good!

PR #553, opened 2024-05-08
Make if predicate passthrough destructively rewrite

Alex-Fischman:
Also deletes the theta passthrough comment because Oliver implemented it on his branch.


oflatt (approved):
Great!

PR #554, opened 2024-05-08
Prettyprinter

clyben:
Add a pretty printer which either output to rust macro or let-binded egglog.
Pretty printer could take a string of egglog output and parse it under current schema to RcExpr, or an RcExpr of program, output the rust macro or let-binded egglog.
Add runmode PrettyPrint and OptimizedPrettyPrint, which output the rust macro form of original program or optimized program. 
To make rust macro look better copy it to rust file and "cargo fmt". Though you need to add ".with_arg_types(..)" and "add_ctx()" to make it runnable.


oflatt (changes requested):
This PR is waiting on a refactor, leveraging the existing pretty printer for s-expressions.

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+    // Type/Assum/BaseType -> intermediate variables
+    symbols: indexmap::IndexMap<Symbols, String>,
+    // intermediate variables about to print
+    log: Vec<String>,
+    // intermediate variable -> Type/Assum/BaseType lookup
+    table: std::collections::BTreeMap<String, SymbolLog>,
+}
+
+#[derive(PartialEq, Eq, Hash)]
+enum Symbols {
```
oflatt, line 25: Can we call this NodeRef or something like that?

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+use std::{collections::HashMap, hash::Hash, rc::Rc, vec};
+
+pub struct PrettyPrinter {
+    pub expr: RcExpr,
+    // Type/Assum/BaseType -> intermediate variables
+    symbols: indexmap::IndexMap<Symbols, String>,
+    // intermediate variables about to print
+    log: Vec<String>,
+    // intermediate variable -> Type/Assum/BaseType lookup
+    table: std::collections::BTreeMap<String, SymbolLog>,
```
oflatt, line 21: Why not HashMap here?

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+                let mut lhs = lhs.as_ref().concat_helper();
+                let mut rhs = rhs.as_ref().concat_helper();
+                lhs.append(&mut rhs);
+                lhs
+            }
+            Expr::Single(expr) => {
+                let expr = Self::to_ast(expr.as_ref());
+                vec![expr]
+            }
+            _ => panic!("Not well formed Concat, expr not in Single"),
```
oflatt, line 538: This isn't correct- a concat could take, for example, a Loop as one of the expressions.

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+}
+
+impl Expr {
+    pub fn pretty(&self) -> String {
+        let (term, termdag) = Rc::new(self.clone()).to_egglog();
+        let expr = termdag.term_to_expr(&term);
+        expr.to_sexp().pretty()
+    }
+
+    fn concat_helper(&self) -> Vec<String> {
```
oflatt, line 526: Change this name to something more descriptive- "gather_concat_children"

dag_in_context/src/pretty_print.rs
oflatt: Add a test for this case

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+            And => "and",
+            Or => "or",
+            PtrAdd => "ptradd",
+            Load => "load",
+            Print => "tprint",
+            Free => "free",
+            FGreaterEq => "f_greater_eq",
+            FGreaterThan => "f_greater_than",
+            FLessEq => "f_less_eq",
+            FLessThan => "f_less_than",
```
oflatt, line 763: are these in ast.rs? Do you need to add them?

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+                .replace(&['(', ')', ' ', ','][..], "") //don't count those char when computing size
+                .len();
+            if fold_when(num_shared, *size) {
+                fold(pp, new_expr)
+            } else {
+                new_expr
+            }
+        };
+
+        fn handle_assum(
```
oflatt, line 359: Your method names need to be more descriptive
Perhaps rename to "refactor_shared_assum"

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+                };
+                pp.log.push(old_assume_binding.clone());
+                pp.table
+                    .insert(old_assume_binding.clone(), SymbolLog::Assumption(new_assum));
+            }
+
+            old_assume_binding
+        }
+
+        fn handle_type(ty: &Type, pp: &mut PrettyPrinter) -> Type {
```
oflatt, line 398: These functions should be top level, not nested inside of `fold_expr`

oflatt: Rename to `refactore_shared_type`

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+                    try_insert_fresh(Symbols::Type(tyin.clone()), tyin.abbrev(), self);
+                    try_insert_fresh(Symbols::Type(tyout.clone()), tyout.abbrev(), self);
+                    self.assign_fresh_var(body);
+                }
+                Expr::Symbolic(_) => panic!("no symbolic should occur when assigning freshvar"),
+            }
+        }
+    }
+
+    fn fold_expr(
```
oflatt, line 330: This method needs a more descriptive name too. How about refactor_shared_expr?

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+            binding,
+        )
+    }
+
+    // turn the Expr to a rust ast macro string.
+    // fold_when: provide a function that decide when to fold the macro to a let binding
+    pub fn to_rust(&mut self, fold_when: &dyn Fn(usize, usize) -> bool, binding: String) -> String {
+        self.assign_fresh_var(&self.expr.clone());
+        let res = self.fold_expr(&self.expr.clone(), fold_when, false);
+        let log = self
```
oflatt, line 186: Why is `log` a memeber of `self`?
It would be cleaner if `fold_expr` returned a log

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+use egglog::{Term, TermDag};
+
+use std::{collections::HashMap, hash::Hash, rc::Rc, vec};
+
+pub struct PrettyPrinter {
+    pub expr: RcExpr,
+    // Type/Assum/BaseType -> intermediate variables
+    symbols: indexmap::IndexMap<Symbols, String>,
+    // intermediate variables about to print
+    log: Vec<String>,
```
oflatt, line 19: Why do you need a log? I think the `table` should give you the order things should be printed if you use an indexmap

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+
+#[derive(PartialEq, Eq, Hash)]
+enum Symbols {
+    Type(schema::Type),
+    Assumption(AssumptionRef),
+    Expr(*const schema::Expr),
+}
+
+#[derive(PartialEq, Eq)]
+enum SymbolLog {
```
oflatt, line 32: Rename to `AstNode`

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+    },
+    schema_helpers::AssumptionRef,
+    to_egglog::TreeToEgglog,
+};
+use egglog::{Term, TermDag};
+
+use std::{collections::HashMap, hash::Hash, rc::Rc, vec};
+
+pub struct PrettyPrinter {
+    pub expr: RcExpr,
```
oflatt, line 15: Why is the `expr` part of the PrettyPrinter? Seems like you pass it in anyways

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+        let mut function_bindings = vec![];
+        let functions = self
+            .functions
+            .clone()
+            .into_iter()
+            .map(|expr| match expr.as_ref() {
+                schema::Expr::Function(name, ..) => {
+                    let binding = format!("fun_{name}_");
+                    function_bindings.push(binding.clone());
+                    pp.new_expr(expr.clone());
```
oflatt, line 80: This code is ugly, I think because prettyprinter has an expr for no reason

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+                    pp.to_egglog_default(binding)
```
oflatt, line 81: pass the expr in here instead

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+        let mut pp = PrettyPrinter::from_expr(self.entry.to_owned());
+        let main = pp.to_egglog_default(main_binding.clone());
+        let mut function_bindings = vec![];
+        let functions = self
+            .functions
+            .clone()
+            .into_iter()
+            .map(|expr| match expr.as_ref() {
+                schema::Expr::Function(name, ..) => {
+                    let binding = format!("fun_{name}_");
```
oflatt, line 78: I think it would be better if `to_egglog_default` returned a fresh name

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+                let symbol = self.table.get(fresh_var).unwrap();
+                let ast = symbol.symbol_log_to_str(true);
+                format!("let {fresh_var} = {ast};")
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
+        log + &format!("\nlet {binding} = {};\n", res.to_ast())
+    }
+
+    fn assign_fresh_var(&mut self, expr: &RcExpr) {
```
oflatt, line 199: What does this method do? Needs a doc comment

oflatt: I think I understand what this method does now- it gives a fresh variable to each node in the AST.
However, you are already traversing the AST in the `fold_expr` method.
You can simply give new variables there instead of here, so you don't need this method

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+                pp.log.push(fresh_var.into());
+                pp.table.insert(fresh_var.into(), SymbolLog::Expr(new_expr));
+            }
+            Expr::Symbolic(fresh_var.into())
+        };
+
+        let num_shared = Rc::strong_count(expr);
+        let fold_or_plain = |pp: &mut PrettyPrinter, new_expr: Expr| {
+            let size = &new_expr
+                .to_string()
```
oflatt, line 349: This `to_string()` call might be really slow, right? I think this is a problem

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+
+    pub fn to_egglog_default(&mut self, binding: String) -> String {
+        self.to_egglog(&|rc, len| (rc > 1 && len > 30) || len > 80, binding)
+    }
+
+    // turn the Expr to a nested egglog with intermediate variables.
+    // fold_when: provide a function that decide when to fold the egglog expression to a let binding
+    pub fn to_egglog(
+        &mut self,
+        fold_when: &dyn Fn(usize, usize) -> bool,
```
oflatt, line 156: We need better documentation of what fold_when is, and what `rc` and `len` are

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,831 @@
...
+        fn handle_type(ty: &Type, pp: &mut PrettyPrinter) -> Type {
+            let ty_str = pp.symbols.get(&Symbols::Type(ty.clone())).unwrap().clone();
+            if !pp.table.contains_key(&ty_str) {
+                pp.log.push(ty_str.clone());
+                pp.table.insert(ty_str.clone(), SymbolLog::Type(ty.clone()));
+            }
+            Type::Symbolic(ty_str)
+        }
+
+        match expr.as_ref() {
```
oflatt, line 407: Can we use some new methods `map_children`, `map_types`, and `map_assumptions` to make this much simpler?

oflatt (changes requested):
This is promising, but we should simplify it before we merge

dag_in_context/src/pretty_print.rs
clyben: because Symbolog contain Expr and Assumption which cannot be hashed

dag_in_context/src/pretty_print.rs
clyben: yes I thought it was added

dag_in_context/src/pretty_print.rs
clyben: mainly bacause Expr cannot be hash so I cannot use a indexmap for table. So need a log to keep track of order or insert.

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,753 @@
...
+            Expr::If(..) => "if".into(),
+            Expr::Switch(..) => "switch".into(),
+            Expr::DoWhile(..) => "dowhile".into(),
+            Expr::Arg(..) => "arg".into(),
+            Expr::Function(name, ..) => "fun_".to_owned() + name,
+            Expr::Symbolic(var) => "symbolic_".to_owned() + var,
+        }
+    }
+
+    pub fn to_ast(&self) -> String {
```
oflatt, line 454: add some newlines in this function

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,753 @@
...
+            }
+            Expr::Empty(..) => "empty()".into(),
+            Expr::Single(expr) => {
+                let expr = expr.to_ast();
+                format!("single({expr})")
+            }
+            Expr::Concat(..) => {
+                let vec = Self::gather_concat_children(self);
+                let inside = vec.join(", ");
+                format!("parallel!({inside})")
```
oflatt, line 505: fix bug with all singles

dag_in_context/src/pretty_print.rs
```diff
@@ -0,0 +1,753 @@
...
+            }
+            Expr::Single(expr) => {
+                let expr = Self::to_ast(expr.as_ref());
+                vec![expr]
+            }
+            _ => vec![self.to_ast()],
+        }
+    }
+
+    pub fn abbrev(&self) -> String {
```
oflatt, line 425: this function will either go away or you can incorporate it to use fresh numbers

oflatt (changes requested):
Very close now!

dag_in_context/src/pretty_print.rs
clyben: I would like make fresh seperate by taking an info.

yihozhang:
action items for this PR after a discussion:
- add caveats that `map_*` may violate RcExpr's invariant.
- document behaviors/limitations of the current pretty printer (inlining parallel!, sharing failure when having different contexts)
- Ignore all assumptions when pretty printing to Rust
  - user will add their own top-level types and assumptions

yihozhang (approved):

PR #558, opened 2024-05-09
Improve perf of testing by increasing sharing, disabling brilift

oflatt:
This PR refactors the testing of benchmarks to increase sharing of work. Now each benchmark is only optimized once instead of twice. Also, the resulting binaries are executed once per treatment instead of twice.
Brillift is still tested on our smaller tests, but is disabled for the benchmarks.

Overall, this makes running tests about twice as fast


Alex-Fischman (approved):
LGTM

PR #561, opened 2024-05-09
Add script to dump llvm/eggccir for benchmarks

rtjoa:
See the output here:
https://github.com/egraphs-good/eggcc/tree/script-output/benchmarks/for_debugging/passing

To regenerate, go to `benchmarks/for_debugging` then:
```
#             input file                         output dir
./dump_one.sh ../passing/bril/core/bitshift.bril passing/bril/core
# all benchmarks:
./dump_all.sh
```


oflatt (approved):
bash 💀 but I approve

PR #562, opened 2024-05-10
Supporting floats

yihozhang:
- [x] introduce the float type at egglog and rust level
- [x] Move `tests/failing/brils/floats` to `passing`.


tests/failing/brils/float/float_args.bril
```diff
@@ -1,4 +0,0 @@
-# ARGS: 3 3.0 -0.3
```
yihozhang, line 1: This `print` takes several arguments. Should we support it?

oflatt: Seems okay to skip

runtime/rt.c
```diff
@@ -0,0 +1,69 @@
+#include <stdio.h>
```
oflatt, line 1: Why is this file in c now? We can't use runtime/src/main.rs?

oflatt (changes requested):

runtime/rt.c
yihozhang: We need both c and rust. The bril repo has the `.rs` runtime for brillvm and `.c` runtime for brilift. See `src/util.rs`.

oflatt: Oh, I see. Thank you

PR #563, opened 2024-05-10
Migrate to fork of Bril and improve runtime installation

oflatt:
This PR gets rid of the ugly part of `install.sh` by forking bril and allowing us to use brillvm as a library.


runtime/build.rs
```diff
@@ -0,0 +1,11 @@
+fn main() {
```
Alex-Fischman, line 1: What does this do?

Alex-Fischman: See https://github.com/sampsyo/bril/pull/319

Alex-Fischman (approved):
LGTM

PR #565, opened 2024-05-10
Add `select` to Bril

ajpal:


Cargo.toml
```diff
@@ -23,11 +23,11 @@ smallvec = "1.11.1"
 syn = { version = "2.0", features = ["full", "extra-traits"] }
 # currently using Pat-Lafon's fork of bril with brillvm fixes
 # on the llvm-18 branch
-bril2json = { git = "https://github.com/Pat-Lafon/bril", rev = "67738d440d3516a99f9cdffb19c91540bfc66c8c" }
-brilirs = { git = "https://github.com/Pat-Lafon/bril", rev = "67738d440d3516a99f9cdffb19c91540bfc66c8c" }
-bril-rs = { git = "https://github.com/Pat-Lafon/bril", rev = "67738d440d3516a99f9cdffb19c91540bfc66c8c" }
-brilift = { git = "https://github.com/Pat-Lafon/bril", rev = "67738d440d3516a99f9cdffb19c91540bfc66c8c" }
-rs2bril = { git = "https://github.com/Pat-Lafon/bril", rev = "67738d440d3516a99f9cdffb19c91540bfc66c8c", features = [
+bril2json = { path = "../bril/bril-rs/bril2json" }
```
oflatt, line 26: Here's our new bril fork:
https://github.com/uwplse/bril/tree/eggcc-main
Could you make a PR there? Thanks!

dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -742,6 +742,7 @@ impl CostModel for DefaultCostModel {
             "Div" => 50.,
             // Comparisons
             "Eq" | "LessThan" | "GreaterThan" | "LessEq" | "GreaterEq" => 10.,
+            "Select" => 10., // @oflatt is this a reasonable value? I made it up...
```
oflatt, line 745: I have no idea but looks good to me haha

dag_in_context/src/interpreter.rs
```diff
@@ -193,6 +193,14 @@ impl<'a> VirtualMachine<'a> {
                 self.memory.insert(pointer.addr(), val);
                 Value::StateV
             }
+            TernaryOp::Select => {
+                let get_bool = |e: &RcExpr, vm: &mut Self| vm.interp_bool_expr(e, arg);
+                if get_bool(e1, self) {
+                    self.interpret_expr(e2, arg)
```
oflatt, line 199: We should think carefully here: do we need to interpret both, then select between the two?
I guess not, since all the effects will still happen due to the state edge? This is probably correct, might be worth a comment

oflatt (approved):
Looks good to me! I'll be interested to see how the bril changes work, we can merge this after that one goes through

PR #566, opened 2024-05-10
Get eggcc repo root from target rather than $EGGCC_ROOT

rtjoa:


oflatt (approved):
Thanks!

oflatt:
We'll revisit this problem and use [Is there any way to include binary or text files in a Rust library?](https://stackoverflow.com/questions/32748918/is-there-any-way-to-include-binary-or-text-files-in-a-rust-library)

PR #571, opened 2024-05-10
Extract function inlining based on expression size

kirstenmg:
The inlining size threshold is passed to `DefaultCostModel` to allow for testing (which I do not actually do).


oflatt:
Closing for now

PR #572, opened 2024-05-13
Add llvm output to nightly

ajpal:
Link in the table:
image

Opens in a new tab:
image

### DIFF
--- a/infra/nightly-resources/handlers.js
+++ b/infra/nightly-resources/handlers.js
@@ -17,6 +17,22 @@ async function load() {
   initializeChart();
 }
 
+// Top-level load function for the llvm page
+async function load_llvm() {
+  GLOBAL_DATA.currentRun = await getBench("./");
+  const params = new URLSearchParams(window.location.search);
+  const benchmark = params.get("benchmark");
+  const runMode = params.get("runmode");
+  if (!benchmark || !runMode) {
+    console.error("missing query params, this probably shouldn't happen");
+  }
+  const llvm = GLOBAL_DATA.currentRun[benchmark][runMode].llvm;
+  if (!llvm) {
+    console.error("missing llvm, this probably shouldn't happen");
+  }
+  document.getElementById("llvm").innerText = llvm;
+}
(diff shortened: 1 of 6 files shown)


infra/nightly-resources/handlers.js
```diff
@@ -17,6 +17,22 @@ async function load() {
   initializeChart();
 }
 
+// Top-level load function for the llvm page
+async function load_llvm() {
+  GLOBAL_DATA.currentRun = await getBench("./");
```
oflatt, line 22: Is `currentRun` used anywhere else? Does it need to be in global data?
If you want to use it in other places in the future we could do this request on startup, perhaps that would make more sense?

infra/profile.py
```diff
@@ -34,16 +56,21 @@ def bench(profile):
...
-    with open(f'{profile_dir}/{name}-args') as f:
+    with open(f'{profile_dir}/{mode}-args') as f:
       args = f.read().rstrip()
     
     # TODO for final nightly results, remove `--max-runs 2` and let hyperfine find stable results
-    subprocess.call(f'hyperfine --warmup 1 --max-runs 2 --export-json {profile_dir}/{name}.json "{profile_dir}/{name} {args}"', shell=True)
+    subprocess.call(f'hyperfine --warmup 1 --max-runs 2 --export-json {profile_dir}/{mode}.json "{profile_dir}/{mode} {args}"', shell=True)
+
+def get_llvm(runMethod, benchmark):
+  files = [f for f in glob(f'./tmp/bench/{benchmark}/llvm-{runMethod}/*') if "init" not in f]
```
oflatt, line 70: I think this is the kind of code that could get us in trouble, especially if we change the naming conventions.

Could you please instead change the code that emits llvm to use the same naming convention as above? e.g. `egglog_noopt_bril_llvm_opt.ll`

oflatt (changes requested):
🎉 This is great!
Just want to avoid nasty bugs with naming in the future

PR #574, opened 2024-05-15
Fix bug in strength reduction

oflatt:
This PR fixed a bug in loop strength reduction. Looks like we mixed up two indices
I also refactored it slightly so that we don't create invalid terms


kirstenmg (approved):
Thank you!

PR #577, opened 2024-05-16
Add live outputs from the loop tail to the loop params

ezrosent:
Added a test that goes from failing to passing with this change. In this case the `ans` variable is overwritten in the loop body but it is still used outside so it needs to be given an input and output.

Note that for variables that are truly unbound at this point we may have to add a placeholder variable to get them working.


tests/passing/small/sqrt.bril
```diff
@@ -0,0 +1,24 @@
+# ARGS: 16
```
yihozhang, line 1: nit: sqrt might be a misleading name since the simplified example isn't doing sqrt

yihozhang (approved):
The fix looks good but gramschmidt.bril is still failing due to non-linearity. Could be related to extraction.

oflatt (approved):
Great!

tests/passing/small/sqrt.bril
ezrosent: Went ahead and made the test more real

PR #579, opened 2024-05-16
Add `select` operator

ajpal:
Going to close https://github.com/egraphs-good/eggcc/pull/565, I'm not sure what exactly is going on in that PR, but one of the other little changes I made in that diff must be slowing down the CI.


oflatt (approved):
Great!

PR #580, opened 2024-05-16
Big ol nightly refactor

ajpal:
Make the data flatt

### DIFF
--- a/infra/nightly-resources/handlers.js
+++ b/infra/nightly-resources/handlers.js
@@ -1,6 +1,7 @@
 // Load data that both the index page and llvm page need
 async function loadCommonData() {
   GLOBAL_DATA.currentRun = await getBench("./");
+  GLOBAL_DATA.currentRunRaw = await getDataJson(".");
 }
 
 // Top-level load function for the main index page.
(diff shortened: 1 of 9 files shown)


infra/nightly-resources/handlers.js
```diff
@@ -74,18 +78,8 @@ function toggleCheckbox(mode, set) {
 }
 
 async function loadBaseline(url) {
-  const data = await getBench(url + "/");
   clearWarnings();
-  GLOBAL_DATA.baselineRun = data;
-  const benchmarkNames = Object.keys(GLOBAL_DATA.currentRun);
-  // Add warnings if the baseline run had a benchmark that the current run doesn't
```
oflatt, line 81: I think this warning got dropped in this PR

infra/nightly-resources/index.js
```diff
@@ -30,254 +30,22 @@ function clearWarnings() {
...
-  parsed.sort((l, r) => {
-    if (l.name < r.name) {
-      return -1;
-    }
-    if (l.name > r.name) {
-      return 1;
-    }
-    return 0;
+  const byBench = {};
+  Array.from(GLOBAL_DATA.enabledBenchmarks).forEach((benchmark) => {
```
oflatt, line 39: Why do we need `Array.from`?

oflatt (changes requested):
Looks good!
However, running it locally it looks like the links to llvm files are broken

infra/nightly-resources/index.js
ajpal: It's a `Set`

ajpal: Actually, not needed, because order of iteration won't matter here.

oflatt (approved):
Great!

PR #581, opened 2024-05-16
Cleaner RVSDG translation using Phi nodes

oflatt:
This PR avoids extra blocks by using `phi` nodes to select between different values.
This uncovered a bug in brillvm, and also our direct jump optimizer had to take into account phi nodes.

### DIFF
--- a/src/rvsdg/optimize_direct_jumps.rs
+++ b/src/rvsdg/optimize_direct_jumps.rs
@@ -151,14 +151,14 @@ impl SimpleCfgFunction {
                 .collect();
             let new_parents = parents
                 .iter()
-                .map(|parent| self.get_parent_skipping_empty(*parent))
+                .map(|parent| self.get_parent_after_collapse(*parent))
                 .collect::<HashSet<_>>();
             let parent_mapping = parents
                 .iter()
                 .map(|parent| {
                     (
                         self.graph[*parent].name.to_string(),
-                        self.graph[self.get_parent_skipping_empty(*parent)]
+                        self.graph[self.get_parent_after_collapse(*parent)]
                             .name
                             .to_string(),
                     )
@@ -177,6 +177,8 @@ impl SimpleCfgFunction {
         self
     }
 
+    /// Detect the case where source -> empty -> target
+    /// The empty block should have a single incoming and single outgoing edge
     fn get_single_in_single_out(&self, parent_block: NodeIndex) -> Option<(EdgeIndex, EdgeIndex)> {
         if !self.graph[parent_block].instrs.is_empty()
             || !self.graph[parent_block].footer.is_empty()
@@ -199,7 +201,9 @@ impl SimpleCfgFunction {
         }
     }
 
-    fn get_parent_skipping_empty(&self, parent_block: NodeIndex) -> NodeIndex {
+    /// Given a parent block, return the block that will be the new parent
+    /// after collapsing the empty block
+    fn get_parent_after_collapse(&self, parent_block: NodeIndex) -> NodeIndex {
         if let Some((source_edge, _)) = self.get_single_in_single_out(parent_block) {
             self.graph.edge_endpoints(source_edge).unwrap().0
         } else {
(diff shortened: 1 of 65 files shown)


Alex-Fischman:
This PR also disables brilift in nightlies.

src/rvsdg/optimize_direct_jumps.rs
```diff
@@ -88,6 +90,24 @@ impl SimpleCfgFunction {
...
+        for (old, new) in node_mapping.iter() {
+            assert!(
+                label_mapping
+                    .insert(
+                        self.graph[*old].name.to_string(),
+                        resulting_graph[*new].name.to_string(),
+                    )
+                    .is_none(),
+                "Duplicate labels in graph"
+            );
```
Alex-Fischman, line 103: nit: I know it's technically correct but stateful expressions in asserts is sus.

src/rvsdg/optimize_direct_jumps.rs
```diff
@@ -99,51 +119,105 @@ impl SimpleCfgFunction {
...
+                    (
+                        self.graph[*parent].name.to_string(),
+                        self.graph[self.get_parent_skipping_empty(*parent)]
+                            .name
+                            .to_string(),
+                    )
+                })
+                .collect::<HashMap<_, _>>();
+            // if we have fewer parents, phi nodes may get messed up, so avoid the optimization
+            if new_parents.len() == parents.len() {
```
Alex-Fischman, line 168: This makes the optimization markedly less effective. If both parents become the same, that means we have a diamond pattern where the sides are empty, and we 100% want to simplify that. You should make `fixup_phis` handle the case where phi instruction become id instructions.

src/rvsdg/optimize_direct_jumps.rs
```diff
@@ -99,51 +119,105 @@ impl SimpleCfgFunction {
...
+        if let ([source_edge], [outgoing]) = (incoming.as_slice(), outgoing.as_slice()) {
+            Some((source_edge.id(), outgoing.id()))
+        } else {
+            None
+        }
+    }
+
+    fn get_parent_skipping_empty(&self, parent_block: NodeIndex) -> NodeIndex {
+        if let Some((source_edge, _)) = self.get_single_in_single_out(parent_block) {
+            self.graph.edge_endpoints(source_edge).unwrap().0
```
Alex-Fischman, line 204: What if the grandparent is also empty? I think this function should be recursive...

Alex-Fischman (changes requested):

src/rvsdg/optimize_direct_jumps.rs
oflatt: fixed

src/rvsdg/optimize_direct_jumps.rs
oflatt: Actually, if both parents are empty your `select` optimization should fire. So this case should be handled by egglog.

src/rvsdg/optimize_direct_jumps.rs
oflatt: Changed the name of the func- it needs to not be recursive for correctness

oflatt:
Oops, re-added branch

src/util.rs
```diff
@@ -471,14 +466,9 @@ impl Run {
...
-                self.optimize_egglog.unwrap(),
-                self.optimize_brilift.unwrap(),
-            ) {
-                (false, false) => "-O0",
-                (true, false) => "-O0-eggcc",
-                (false, true) => "-O3",
-                (true, true) => "-O3-eggcc",
+            name += match self.optimize_brilift.unwrap() {
+                false => "-O0",
+                true => "-O3",
```
Alex-Fischman, line 471: The brilift runmode comments that I wrote are wrong now, you should fix them.

Alex-Fischman (approved):
Approved once comments are fixed

PR #588, opened 2024-05-17
Function inlining: always inline for fixed iterations

kirstenmg:
Change function inlining to always inline for our existing fixed number of iterations. 
This does cause code size to increase for recursive functions, but it works for now.


dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -96,12 +99,35 @@ impl<'a> EgraphInfo<'a> {
         loop_iteration_estimates
     }
 
+    fn get_inlined_calls(egraph: &EGraph) -> HashSet<(ClassId, ClassId)> {
+        let mut inlined_calls = HashSet::new();
+
+        // loop over all nodes, finding DontExtract nodes
```
oflatt, line 105: Comment is out of date

dag_in_context/src/lib.rs
```diff
@@ -113,28 +113,56 @@ pub(crate) fn print_with_intermediate_vars(termdag: &TermDag, term: Term) -> Str
...
+                    term_cache,
+                    printed,
+                );
+                let body_term = cb.body.to_egglog_internal(tree_state);
+                format!(
+                    // (union call body) (InlinedCall "callee" args) (subsume (Call "callee" args))
+                    // We need to subsume, otherwise the Call in the original program could get
+                    // substituted into another context during optimization and no longer match InlinedCall.
+                    "(union {} {})\n(InlinedCall \"{}\" {})\n(subsume (Call \"{}\" {}))\n",
+                    print_with_intermediate_helper(
```
oflatt, line 142: This can also be (Call {} {}), simplifying this code.
I suggest binding two variables:
`call_with_args` and `inlined`, and use them here.

dag_in_context/src/lib.rs
```diff
@@ -113,28 +113,56 @@ pub(crate) fn print_with_intermediate_vars(termdag: &TermDag, term: Term) -> Str
...
+                    call_args.clone(),
+                    term_cache,
+                    printed,
+                );
+                let body_term = cb.body.to_egglog_internal(tree_state);
+                format!(
+                    // (union call body) (InlinedCall "callee" args) (subsume (Call "callee" args))
+                    // We need to subsume, otherwise the Call in the original program could get
+                    // substituted into another context during optimization and no longer match InlinedCall.
+                    "(union {} {})\n(InlinedCall \"{}\" {})\n(subsume (Call \"{}\" {}))\n",
```
oflatt, line 141: Use rust newlines instead of `\n`

dag_in_context/src/lib.rs
```diff
@@ -113,28 +113,56 @@ pub(crate) fn print_with_intermediate_vars(termdag: &TermDag, term: Term) -> Str
...
-#[allow(dead_code)]
 fn print_function_inlining_pairs(
     function_inlining_pairs: Vec<function_inlining::CallBody>,
     printed: &mut String,
     tree_state: &mut TreeToEgglog,
     term_cache: &mut HashMap<Term, String>,
 ) -> String {
-    // Get unions
+    // Get unions and mark each call as inlined for extraction purposes
+    let inlined_call = "(relation InlinedCall (String Expr))";
```
oflatt, line 123: can you put this somewhere in a .egg file, along with the loop estimate table?

oflatt (changes requested):
Great! Some refactor

dag_in_context/src/lib.rs
kirstenmg: I think it might be better to keep it here, since then it is by its usage. It shouldn't be used for any other optimization. `LoopNumItersGuess`, for example, is in the same .egg file as its usage.

kirstenmg: I do think it would be better to put this whole function into function_inlining.rs, though.

oflatt (approved):
Looks great!

PR #590, opened 2024-05-18
get rid of dependency on filesystem

ryan-berger:


ryan-berger:
@oflatt @ajpal we do need to wait for the nightly to run tonight, but it works well on my machine (which is nearly identical to the nightly box)

ryan-berger:
@oflatt Nightlies seem to be broken actually. `benchmarks/passing/bril/float/conjugate-gradient.ll`. But looks like it happens on main too

oflatt (approved):
This looks great!
I see that the chart is broken in main, and nightly is broken too. Lets hold off on merging, we'll fix it ASAP

oflatt:
Works!

PR #592, opened 2024-05-20
Fix frontend crash error

ajpal:
We were seeing this error:
image
If there was no previous run to compare to

Turns out, there wasn't a `main` run in the last 30 runs, so the slice of previous runs we returned didn't have a `main` run.
Fix is to make sure we have at least on `main` run in the slice.

### DIFF
--- a/infra/profile.py
+++ b/infra/profile.py
@@ -92,8 +92,11 @@ def bench(benchmark):
 def get_llvm(runMethod, benchmark):
   path = f'./tmp/bench/{benchmark}/llvm-{runMethod}/{benchmark}-{runMethod}.ll'
 
-  with open(path) as f:
-    return f.read()
+  try:
+    with open(path) as f:
+      return f.read()
+  except OSError:
+    return ""
 
 
 # aggregate all profile info into a single json array.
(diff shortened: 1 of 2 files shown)


infra/profile.py
```diff
@@ -92,8 +92,11 @@ def bench(benchmark):
 def get_llvm(runMethod, benchmark):
   path = f'./tmp/bench/{benchmark}/llvm-{runMethod}/{benchmark}-{runMethod}.ll'
 
-  with open(path) as f:
-    return f.read()
+  try:
```
oflatt, line 95: Shouldn't we return an error here or something?

oflatt (changes requested):
Good fix! Concerned about error reporting on .ll files

infra/profile.py
ajpal: I think it's better to handle errors gracefully when possible- I wouldn't want the whole nightly run to crash if one benchmark doesn't output the llvm file. I think it's better to return `''` which we can detect on the frontend and show like "hey, this benchmark is missing the llvm"

oflatt: Ah yes I agree, I just want to make sure we write the code that detects the missing llvm and throws an error

PR #593, opened 2024-05-20
Add warning for missing llvm and render table latex

ajpal:
image

@ryan-berger I don't think the latex is correct here.. When I copy it into overleaf I get all kinds of errors:
image


oflatt (approved):
Thanks!

PR #594, opened 2024-05-20
Run mem2reg in the baseline, since brillvm relies on it

oflatt:


oflatt:
Oh right! I got confused, I've been trying to minimize the flags we send to opt

PR #597, opened 2024-05-21
Show baseline as dotted line at 1 for speedup

ajpal:
image

Also puts a scrollbox on the benchmarks so it's more manageable:
image

### DIFF
--- a/infra/nightly-resources/chart.js
+++ b/infra/nightly-resources/chart.js
@@ -32,7 +32,10 @@ function getValue(entry) {
     if (!baseline) {
       addWarning(`No speedup baseline for ${benchmark}`);
     }
-    return baseline.hyperfine.results[0].mean / entry.hyperfine.results[0].mean;
+    const baseV = baseline.hyperfine.results[0].mean;
+    const expV = entry.hyperfine.results[0].mean;
+    // return (baseV - expV) / expV * 100;
+    return baseV / expV;
   } else {
     throw new Error(`unknown chart mode ${GLOBAL_DATA.chart.mode}`);
   }
@@ -81,7 +84,7 @@ function parseDataForChart() {
     datasets[mode] = {
       label: mode,
       backgroundColor: COLORS[mode],
-      data: Array(benchmarks.length).fill(0),
+      data: Array(benchmarks.size).fill(0),
       borderWidth: 1,
       errorBars: {},
     };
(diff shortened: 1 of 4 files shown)


infra/nightly-resources/chart.js
```diff
@@ -81,7 +84,7 @@ function parseDataForChart() {
     datasets[mode] = {
       label: mode,
       backgroundColor: COLORS[mode],
-      data: Array(benchmarks.length).fill(0),
+      data: Array(benchmarks.size).fill(0),
```
oflatt, line 87: did you mean benchmarks.size here? Idk the difference but a cursory google search says length is right for arrays

infra/nightly-resources/chart.js
```diff
@@ -32,7 +32,10 @@ function getValue(entry) {
     if (!baseline) {
       addWarning(`No speedup baseline for ${benchmark}`);
     }
-    return baseline.hyperfine.results[0].mean / entry.hyperfine.results[0].mean;
+    const baseV = baseline.hyperfine.results[0].mean;
+    const expV = entry.hyperfine.results[0].mean;
+    // return (baseV - expV) / expV * 100;
+    return baseV / expV;
```
oflatt, line 38: would be great to show this formula on the screen (another PR is good)

infra/nightly-resources/chart.js
```diff
@@ -112,11 +131,6 @@ function initializeChart() {
     type: "bar",
     data: chartData,
     options: {
-      scales: {
-        y: {
-          beginAtZero: true,
```
oflatt, line 117: is this bc it wasn't working before?

infra/nightly-resources/chart.js
```diff
@@ -97,6 +100,22 @@ function parseDataForChart() {
...
     });
   });
+
+  // Show baseline as dotted line at 1x if speedup
+  if (GLOBAL_DATA.chart.mode === "speedup") {
+    datasets[BASELINE_MODE] = {
+      label: BASELINE_MODE,
+      data: Array(benchmarks.size + 1).fill(1),
+      type: "line",
+      borderColor: "purple",
```
oflatt, line 110: perhaps this should be colors[baseline]

oflatt (approved):
Looks great! None of my comments are blocking

infra/nightly-resources/chart.js
ajpal: Yeah, it's a set not an array

infra/nightly-resources/chart.js
ajpal: Yeah I copied some of this code from the ChartJS docs, but I don't think this was actually doing anything

PR #602, opened 2024-05-21
Optimize if statements that pass through one side

oflatt:


src/rvsdg/to_cfg.rs
```diff
@@ -461,8 +481,18 @@ impl<'a> RvsdgToCfg<'a> {
...
-                    branch_blocks
-                        .push(self.sequence_results(&[outputs_for_branch, output_assigned]));
+                    if first_passthrough == Some(i) {
+                        // Optimization: if this is the passthrough block, run it first before all other branches
+                        before_if = self.sequence_results(&[
+                            before_if,
+                            outputs_for_branch,
+                            output_assigned,
+                        ]);
+                        branch_blocks.push(self.sequence_results(&[]));
```
ezrosent, line 491: So, to clarify we are still creating a block for the passthrough case, but now `optimize_direct_jumps` is going to clean that block up?

ezrosent (approved):
Looks good! Diff is mostly snapshots. Just one clarification question.

Alex-Fischman (approved):
LGTM

src/rvsdg/to_cfg.rs
oflatt: yes!

PR #604, opened 2024-05-22
Store LLVM separately on nightly

ajpal:
The latest nightly ([link](https://nightly.cs.washington.edu/reports/eggcc/1716340359:nightly:main:30dbf39f86/)) is broken because the profile.json is huge (half a gig!)

To mitigate, we should store the LLVM separately for each run mode and benchmark
This has the added benefit that we can periodically go delete these files from nightly without losing the rest of the profile info


oflatt (approved):
Looks good!

Issue #457, opened 2024-04-15
Theta Passthrough Rule


Alex-Fischman:
Right now the theta passthrough rule is commented out because it's unsound when the loop doesn't terminate. Fix this by only firing the rule when we know that the loop terminates.

PR #607, opened 2024-05-23
[Depends on #606] Add state edge passthrough for terminating loops

kirstenmg:
Moved loop iteration analysis into its own file; allowed state edge passthrough of loops that are confirmed to terminate. It's not visible in the snapshots, but if you look at the optimized RVSDG for `loop_pass_through.bril` you can see the state edge moved out.


Alex-Fischman:
Closes #457.

oflatt:
This is old

PR #608, opened 2024-05-23
Show CFGs on Nightly

ajpal:
image

image
image

### DIFF
--- a/infra/generate_cfgs.py
+++ b/infra/generate_cfgs.py
@@ -3,11 +3,9 @@ import glob
 import os
 
 def make_cfgs(bench, data_dir):
-  print(bench)
   cwd = os.getcwd()
   path = f"{data_dir}/{bench}"
   runmodes = os.listdir(path)
-  print(runmodes)
   for mode in runmodes:
     os.chdir(f"{path}/{mode}")
 
@@ -26,6 +24,7 @@ def make_cfgs(bench, data_dir):
       os.system(cmd)
 
     pngs = glob.glob("*.png")
+    print(f"Generated {len(pngs)} CFGs for {bench} {mode}")
     with open("png_names.txt", "w") as f:
       f.write("\n".join(pngs))
 
@@ -42,7 +41,6 @@ if __name__ == '__main__':
       print("Usage: generate_line_counts.py <data directory>")
       exit(1)
   data_dir = os.sys.argv[1]
-  print(data_dir)
   benchmarks = os.listdir(data_dir)
   for bench in benchmarks:
     make_cfgs(bench, data_dir)
--- a/infra/generate_cfgs.py
+++ b/infra/generate_cfgs.py
@@ -9,8 +9,17 @@ def make_cfgs(bench, data_dir):
   for mode in runmodes:
     os.chdir(f"{path}/{mode}")
 
+    # HACK: check if opt-18 exists
+    # otherwise use opt
+    # On Linux, sometimes it's called opt-18, while on mac it seems to be just opt
+    # Also, on some machines, just running `opt-18` hangs, so we pass the version flag
+    if os.system("opt-18 --version") == 0:
+      opt = "opt-18"
+    else:
+      opt = "opt"
+
     # https://llvm.org/docs/Passes.html#dot-cfg-print-cfg-of-function-to-dot-file
-    cmd = "opt -disable-output -passes=dot-cfg llvm.ll > /dev/null 2>&1"
+    cmd = f"{opt} -disable-output -passes=dot-cfg llvm.ll"
     os.system(cmd)
 
     # Find all the dot files (can't use glob because it doesn't match hidden files)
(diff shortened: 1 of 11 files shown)


infra/nightly.sh
```diff
@@ -62,6 +62,9 @@ fi
 # Generate latex after running the profiler (depends on profile.json)
 ./infra/generate_line_counts.py "$NIGHTLY_DIR"
 
+# Generate CFGs for LLVM after running the profiler
+./infra/generate_cfgs.py "$NIGHTLY_DIR/data/llvm"
```
oflatt, line 66: Might be better for the `profile.py` file to call `generate_cfgs` directly- less bash communication is better

infra/profile.py
```diff
@@ -103,8 +103,9 @@ def should_have_llvm_ir(runMethod):
 def dump_llvm_ir(runMethod, benchmark, output_directory):
```
oflatt, line 103: What is the point of `dump_llvm_ir`? It seems to move .ll files from one folder to another. Shouldn't we just put it in that folder in the first place?

oflatt (changes requested):
This code requires lots of mutating the file system, so I'm getting confused about all the folders. If we can minimize the number of places we specify paths and minimize moving files around that would be great.

Besides that, making clear wrappers for mutating the state of the html page would be good

infra/nightly.sh
ajpal: I kind of disagree? I think the separation is good- `profile.py` is in charge of running all the benchmarks and stuff, `generate_cfgs` is in charge of reading that data in and writing pngs out.

infra/nightly-resources/handlers.js
```diff
@@ -99,39 +100,45 @@ async function loadBaseline(url) {
...
-  GLOBAL_DATA.chart.mode = elt.value;
-  document.getElementById("speedup-formula").style.visibility =
-    elt.value === "speedup" ? "visible" : "hidden";
-  refreshChart();
-}
-
-function copyToClipboard(eltId) {
-  navigator.clipboard.writeText(document.getElementById(eltId).innerText);
+function toggleAllPngs(elt) {
+  const btns = Array.from(document.getElementsByTagName("button"));
```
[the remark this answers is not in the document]
ajpal: Changed to getting elements by class name

infra/nightly-resources/llvm.js
```diff
@@ -0,0 +1,43 @@
+async function showIR(benchmark, runMode) {
+  const llvm = await fetchText(`./data/llvm/${benchmark}/${runMode}/llvm.ll`);
+  document.getElementById("llvm-ir").innerText = llvm;
+}
+
+async function showCFGs(benchmark, runMode) {
+  let pngs = (
+    await fetchText(`./data/llvm/${benchmark}/${runMode}/png_names.txt`)
```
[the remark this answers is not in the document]
ajpal: Discussed offline- this is the least bad option :)

infra/profile.py
ajpal: Fixed

infra/generate_cfgs.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+import glob
+import os
+
+def make_cfgs(bench, data_dir):
+  cwd = os.getcwd()
+  path = f"{data_dir}/{bench}"
+  runmodes = os.listdir(path)
+  for mode in runmodes:
+    os.chdir(f"{path}/{mode}")
```
oflatt, line 10: Sorry I didn't catch this, but I think it's bad practice to change the working directory
Can we build the relative path everywhere instead?

infra/generate_cfgs.py
```diff
@@ -0,0 +1,59 @@
...
+    else:
+      opt = "opt"
+
+    # https://llvm.org/docs/Passes.html#dot-cfg-print-cfg-of-function-to-dot-file
+    cmd = f"{opt} -disable-output -passes=dot-cfg {bench}-{mode}.ll"
+    os.system(cmd)
+
+    # Delete the -init.ll file (We don't need it for nightly,
+    # so just reduce the amount of clutter we copy to the nightly machine)
+    os.system(f"rm {bench}-{mode}-init.ll")
```
oflatt, line 27: Should this be deleted when it is created instead of here?

infra/profile.py
```diff
@@ -129,17 +123,19 @@ def aggregate(compile_times, bench_times, output_directory):
     print("Usage: profile.py <bril_directory> <output_directory>")
     exit(1)
 
-  profile_path, output_path = os.sys.argv[1:]
+  # Create tmp directory for intermediate files
+  os.mkdir(TMP_DIR)
```
oflatt, line 127: If this script is killed and the tmp dir is still there, does this fail? Do we want to delete it in that case?

oflatt (approved):
Looks great! Minor comments, not blocking

infra/generate_cfgs.py
ajpal: Then all the dot and png files would get created in the top-level dir, which is not what we want

PR #610, opened 2024-05-26
Simplify control flow for CFGs generated by eggcc.

ezrosent:
There's a fairly detailed doc comment in `simplify_branches.rs`. The goal here is to clean up the CFG output and remove redundant branches from the generated code. Here's an example: at head, the `sqrt` benchmark generates the following cfg:

sqrt-old

And with this PR, we get:

sqrt-new


ezrosent:
Question for reviewers: this will remove more branches if we iterate to a fixed point. It's easy to add; let me know if you want to try it.

ezrosent:
Went ahead and added that in the next commit. Figure it's worth looking at nightly with everything. It's a small change and easy to revert. Here's the result for the same sqrt program.
sqrt-fp

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+#[derive(Clone, Default, Debug)]
+enum ValueInfo {
+    #[default]
+    Bot,
+    Known(Literal),
+    Top,
+}
+
+impl ValueInfo {
+    fn merge(&mut self, other: &ValueInfo) -> bool {
```
oflatt, line 370: Comment: returns whether anything has changed

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+        // Not handling multi-way branches for now.
+        None
+    }
+}
+
+/// A basic semilattice describing the state of a value.
+#[derive(Clone, Default, Debug)]
+enum ValueInfo {
+    #[default]
+    Bot,
```
oflatt, line 364: Worth commenting intuition about `Bot` and `Top`. I think `Bot` is "knows nothing" and `Top` is unreachable because we found a conflict

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+    }
+}
+
+/// The state of the (boolean) values in a particular basic block.
+#[derive(Default, Debug)]
+struct ValueState {
+    /// The (pointwise) join of all of the values in incoming branches.
+    inherited: IndexMap<Identifier, ValueInfo>,
+    /// The transforms induced by any operations on variables in the block.
+    transforms: VecDeque<(
```
oflatt, line 421: comment: computed once, on initialization. Then virtual assignments are added to it during step 2

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+        Identifier, /* dst  */
+        Identifier, /* src */
+        Transform,
+    )>,
+    /// The set of variables written to in this basic block.
+    kills: HashSet<Identifier>,
```
oflatt, line 427: As far as I can tell this is the variables written to in `transforms`. Could we just initialize it after step 2?

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+            // * For each outgoing branch [edge e1] with cond val `v` for `id`
+            // * Check if `id` was written to in `n`, if it was, then move on
+            // _unless_ we know the value of `id`; in which case we can replace the branch with a jump.
+            // * Otherwise, check if a predecessor [via edge e2] node has `v` as a
+            // known value for `id`.
+            // * If so, copy the contents of the admin node to that predecessor, and
+            // reroute e2 to the target of e1.
+            let mut scratch = Vec::new();
+            for admin_node in &branch_meta.admin_nodes {
+                let mut walker = self
```
oflatt, line 134: Could you break this out into a helper?

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+                    .graph
+                    .neighbors_directed(*admin_node, Direction::Outgoing)
+                    .detach();
+                // Don't reroute past the exit node. We want to make sure it stays reachable.
+                if admin_node == &self.exit {
+                    continue;
+                }
+                while let Some((outgoing, succ)) = walker.next(&self.graph) {
+                    let BranchOp::Cond { arg, val, .. } = self.graph[outgoing].op.clone() else {
+                        continue;
```
oflatt, line 144: All the continues and breaks make this code a little brittle to refactor
Is there a way to combine some of the conditions into boolean / matches! statements?

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+                    };
+                    let Some(val) = to_lit(&val) else {
+                        continue;
```
oflatt, line 147: If we keep the `continue` and `break` statements, might be worth labeling the while loops

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+                    };
+                    if val_analysis.data[admin_node].kills.contains(&arg) {
```
oflatt, line 149: I think this `if` is a special, easy case- can it be broken out into a helper? Or a whole different pass perhaps?

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+                        changed = true;
+                        // Don't run the rest of the inner loop.
+                        break;
+                    }
+                    let mut incoming_walker = self
+                        .graph
+                        .neighbors_directed(*admin_node, Direction::Incoming)
+                        .detach();
+                    while let Some((incoming, pred)) = incoming_walker.next(&self.graph) {
+                        let can_reroute = matches!(val_analysis.data[&pred].get_output(&arg), ValueInfo::Known(v) if v == val);
```
oflatt, line 191: unnecessary assignment

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+                        if !can_reroute {
+                            continue;
+                        }
+
+                        let weight = self.graph.remove_edge(incoming).unwrap();
+                        // We only have to worry about `instrs` because we
+                        // checked that the footer was empty when we populated
+                        // admin_nodes. We do this because we more or less don't
+                        // use footers on our way back to bril.
+                        scratch.extend(self.graph[*admin_node].instrs.iter().cloned());
```
oflatt, line 201: Not sure I understand how `scratch` is getting used. Can `scratch` be a local definition here? What about just above this while loop instead of far up?

src/rvsdg/simplify_branches.rs
```diff
@@ -0,0 +1,623 @@
...
+                            scratch.append(&mut target_block.instrs);
+                            mem::swap(&mut target_block.instrs, &mut scratch);
+                            self.graph.add_edge(pred, target, weight);
+                            changed = true;
+                            break;
+                        } else {
+                            scratch.clear();
+                            // Otherwise we may need some sort of compatibility check to
+                            // merge the block somewhere. Add the edge back for now:
+                            self.graph.add_edge(*admin_node, target, weight);
```
oflatt, line 231: I'd rather we didn't remove the edge in the first place for this one

oflatt (changes requested):
Really great job, excited that this is working!
I think some of the gnarly checks that actually do the graph rewrites can be cleaned up. The actual analysis seems pretty clean.

src/rvsdg/simplify_branches.rs
ezrosent: We can, but then if we added  more calls to `add_assignment` later, we'd then have to remember to recompute this. I think it's a bit less brittle to keep something like the current structure.

src/rvsdg/simplify_branches.rs
ezrosent: In what way is it brittle? The continues and breaks are just a way of doing a filter here, and I think there's basically only one order in which to apply them all (e.g. if we refactored how condition values were represented, the right things in this code will fail to typecheck).

src/rvsdg/simplify_branches.rs
ezrosent: scratch is being used to save repeated allocations when running the algorithm, so the higher up the better I think.

I've moved it inside `rewrite_branches` though.

src/rvsdg/simplify_branches.rs
ezrosent: The thing that makes that harder is the `break` vs. `continue` case. We can lift that into its own type, but that + a bunch of params creates a bit of an odd helper. I think this might be better but I'm not sure.

As for a separate pass I think that'd require that we redo a bunch of the scaffolding for the while loops? Not sure if that's worth it either.

```rust
    enum Next { Break, Continue }
    fn fold_constant_branch(
        &mut self,
        admin_node: NodeIndex,
        succ: NodeIndex,
        arg: &Identifier,
        val_analysis: &ValueAnalysis,
        val: &Literal,
    ) -> (Next, bool) {
        if succ != self.exit && self.graph.neighbors(admin_node).any(|x| x == self.exit) {
            // Don't remove any outgoing links to the exit node.
            return (Next::Break, false);
        }
        // We assign to the branched-on argument in the admin
        // node. See if we can fold the constant branch here.
        let ValueInfo::Known(lit) = val_analysis.data[&admin_node].get_output(arg) else {
            return (Next::Continue, false);
        };
        if &lit != val {
            return (Next::Continue, false);
        }
        // okay, we have found a matching edge. Replace this branch
        // with a jump.
        let mut walker = self
            .graph
            .neighbors_directed(admin_node, Direction::Outgoing)
            .detach();
        while let Some((outgoing, _)) = walker.next(&self.graph) {
            self.graph.remove_edge(outgoing);
        }
        self.graph.add_edge(
            admin_node,
            succ,
            Branch {
                op: BranchOp::Jmp,
                pos: None,
            },
        );
        // Don't run the rest of the inner loop.
        (Next::Break, true)
}
```

src/rvsdg/simplify_branches.rs
ezrosent: When I see labelled while loops I generally assume that some code is going to do a multi-level break... I think the continues and breaks here are pretty clear? Are we doing this elsewhere in the code?

src/rvsdg/simplify_branches.rs
oflatt: It's brittle to refactors, maybe it's a me thing but I often introduce bugs when I change code with continues in it.
If it's the only way that's fine

src/rvsdg/simplify_branches.rs
oflatt: I agree the helper is weird
I still think a separate while loop may be cleaner but I'm not hung up on this

src/rvsdg/simplify_branches.rs
oflatt: Oh I see.
I think `scratch` is a bad name, and I don't care about the repeated allocations.

oflatt (approved):
Really great improvement! This looks ready to merge, it's challenging to refactor further as we've seen from the threads

PR #611, opened 2024-05-28
Add bril ops

ajpal:


dag_in_context/src/greedy_dag_extractor.rs
```diff
@@ -857,15 +857,15 @@ impl CostModel for DefaultCostModel {
             "Base" | "TupleT" | "TNil" | "TCons" => 0.,
             "Int" | "Bool" | "Float" => 0.,
             // Algebra
-            "Add" | "PtrAdd" | "Sub" | "And" | "Or" | "Not" => 10.,
-            "FAdd" | "FSub" => 50.,
+            "Add" | "PtrAdd" | "Sub" | "And" | "Or" | "Not" | "Shl" | "Shr" => 10.,
+            "FAdd" | "FSub" | "Fmax" | "Fmin" => 50.,
```
oflatt, line 861: If you have any justification for these costs it's worth a comment. Otherwise, it's fine we can play with these

oflatt (approved):
Awesome!
It's probably worth adding some bril tests that use the new ops- that way it checks automatically that type checking is working, ect

PR #617, opened 2024-05-29
Using loop-inversion to optimizes translation overhead for while loops

yihozhang:
Prerequisite: #620 

Before:
image

After:
image

There are still some issues that need to be fixed.

Updates: the tests are still blowing up


yihozhang:
Closing since IVT is in main now

PR #618, opened 2024-05-30
Rewrite if a </> b then a else b to min/max

ajpal:
Doesn't result in big changes to benchmarks yet because of the branch thing that @Alex-Fischman is working on.

But I tested on the commit that Alex sent me, and `min.bril` correctly optimizes to
```
@main(v0: int, v1: int) {
.b2_:
  v3_: int = smin v0 v1;
  print v3_;
}
```


oflatt (approved):
Neat!

Alex-Fischman (approved):
Rules look good but I'm not sure they're in the right ruleset. They should perhaps go in `always-switch-rewrites` or in `passthrough`.

PR #620, opened 2024-05-30
Fix ternary operator rules

yihozhang:


oflatt (approved):
Great! Really reveals the need for better testing

PR #622, opened 2024-06-04
Fix state passthrough (banish the shadow realm)

Alex-Fischman:
This PR fixes the state edge passthrough performance bug by "tying the knot" so that loop contexts point to the loop itself (instead of to the pre-context version of the loop). To this end, it also:

- Refactors `add_context.rs`
- Refactors the schedule
- Changes the code that builds the program string in `dag_in_context/src/lib.rs`
- Touches almost every test in the `dag_in_context` subcrate to add the unions to the `egglog_test`
- Because we need to create temporary variables in `add_context.rs`, and because substitution adds context, we have to pass a `ContextCache` throughout our code.
- Deletes the empty `fuction_inlining.egg` file
- Fixes the redundant switch rewrite rules and moves them to passthrough
- Improves the `debug-helper.egg` file
- Does some other stuff too, like adding two new snapshot tests


dag_in_context/Cargo.lock
```diff
@@ -152,7 +152,7 @@ dependencies = [
 [[package]]
 name = "bril-rs"
 version = "0.1.0"
-source = "git+https://github.com/uwplse/bril?rev=78881c45aa53231915f333d1d6dcc26cedc63b57#78881c45aa53231915f333d1d6dcc26cedc63b57"
+source = "git+https://github.com/uwplse/bril?rev=e2be3f5#e2be3f5d7e160f02b7aed0ef2bcc3e13ae722d2b"
```
ajpal, line 155: Why is there a change to the lock but not the toml?

dag_in_context/src/add_context.rs
```diff
@@ -3,25 +3,37 @@
...
     schema_helpers::AssumptionRef,
+    to_egglog::TreeToEgglog,
 };
 
-struct ContextCache {
+pub struct ContextCache {
     with_ctx: HashMap<(*const Expr, AssumptionRef), RcExpr>,
     symbol_gen: HashMap<(*const Expr, AssumptionRef), String>,
+    /// Information for generating the unions for loop contexts
+    placeholder: usize,
```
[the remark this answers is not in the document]
ajpal: edit after reading through the rest of the code:
Here's how I think this is working (but please correct me if I'm wrong!!)
When we add context to a loop, first we create a new context (I'm not sure why it's an `InFunc` assumption) and use that to add context to the pred and body.
Now we have two contexts: the Loop ctx, and the dummy loop context that we made. We add this pair to `loop_context_unions` so that we can add egglog code to union them later. This way, the loop with context doesn't use the original `pred_and_body` term at all (this was the shadow realm).
This is in contrast to the old way of doing it where the loop assumption would point to the `pred_and_body` without context nodes.

Ok, so hopefully that's at least mostly right. If so, I think some of the names could be improved. For example, it's not really clear what `placeholder` is for. I think it's like a counter kind of thing to make sure we have unique names for each of these loop ctx things. I think the name should allude to that (`idx`, `ctr`, something like that). Also, you sometimes re-use the name `placeholder` to refer to the ctx itself. I think it'd be better to not overload the term in that way.

dag_in_context/src/add_context.rs
```diff
@@ -143,12 +228,16 @@ impl Expr {
             Expr::Arg(ty, _oldctx) => RcExpr::new(Expr::Arg(ty.clone(), context_to_add)),
             // create new contexts for let, loop, and if
             Expr::DoWhile(inputs, pred_and_body) => {
+                let placeholder = cache.new_placeholder();
```
ajpal, line 231: I think this code-block could really use a comment (I guess basically the summary I put in a comment above). I think it will be important to capture *why* we need to do all this, especially since the way we had it before seems simpler, and if we later forget all the shadow realm problems, we might accidentally reintroduce those problems.

tests/snapshots/files__block-diamond-optimize.snap
```diff
@@ -10,28 +10,24 @@ expression: visualization.result
   c5_: int = const 0;
```
ajpal, line 10: I didn't look closely at any of these snapshot diffs. Should I ?

dag_in_context/src/add_context.rs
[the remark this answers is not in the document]
Alex-Fischman: Yep, that's all correct. I'll change the name and add a comment to the place where I used `InFunc` (it's a hack to avoid changing the schema).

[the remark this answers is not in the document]
Alex-Fischman: I'm going to use placeholder everywhere to refer to the "dummy" loop context that we will union later. (The problem is that "dummy" contexts already mean something else.)

dag_in_context/src/type_analysis.egg
```diff
@@ -33,7 +33,7 @@
 (rule (
         (ExpectType e expected msg)
         (HasType e actual)
-        (!= expected actual) ;; OKAY to compare types for equality because we never union types.
+        (!= expected actual) ;; not okay unless we saturate type helpers.
```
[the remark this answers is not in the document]
Alex-Fischman: The original comment is wrong; our helper functions do union themselves with types.

dag_in_context/Cargo.lock
Alex-Fischman: Not sure, I think this happened when I rebased on main.

tests/snapshots/files__block-diamond-optimize.snap
Alex-Fischman: Probably not. I skimmed them to make sure that this PR doesn't add a ton of code for some reason, but I see lots of the small examples getting their if statements deleted so I'm pretty sure it's working.

dag_in_context/Cargo.lock
ajpal: I'm pretty sure it's not safe to change one without the other. But looking at `main`, looks like this change is already there: https://github.com/egraphs-good/eggcc/blob/main/Cargo.lock#L199 so you should be okay to discard this?

ajpal:
I'd like to see a solid nightly run before we merge, but code lgtm!

dag_in_context/Cargo.lock
Alex-Fischman: See Yihong's thread on Slack?
I can't discard this change because this file gets rebuilt by Rust.

Alex-Fischman:
I just looked at eggcc-notify and the nightly for main has been failing since May 31st. Since I rebased my branch is also failing. I don't know why.

Alex-Fischman:
Rebased on `fix-install` (#623).

ajpal (approved):

PR #623, opened 2024-06-06
Update install.sh

ajpal:


ajpal:
@Alex-Fischman I think this will fix the error on nightly for both `main` and your branch, so you should pull this change into your branch and re-run

PR #624, opened 2024-06-12
[Nightly] Small fixes

ajpal:
1. Fix expand/collapse for llvm IR
2. link to bril (on github) for each benchmark


infra/nightly-resources/table.js
```diff
@@ -121,42 +121,37 @@ function ConvertJsonToTable(
             var value = parsedJson[i][headers[j]];
             var isUrl = urlRegExp.test(value) || javascriptRegExp.test(value);
 
-            if (isUrl)
```
ajpal, line 124: We don't want this handling for urls (we do it ourselves in index.js)

chandrakananandi (approved):
thanks so much!!

PR #628, opened 2024-08-09
Mechanized semantics and proofs

rtjoa:
Add a mechanized formal semantics for eggcc IR*, as well as proofs of:
- If context introduction
  - `(If pred in then else)` is equivalent to `(If pred in (Ctx (InThen pred in) then) (Ctx (InElse pred in) else)`
- Context pushdown
  - `(Ctx a (PureOp e))` is equivalent to `(PureOp (Ctx a e))`
  - `(Ctx a (If pred in then else))` is equivalent to `(If (Ctx a pred) (Ctx a in) then else)`
  - `(Ctx a (DoWhile in pred out))` is equivalent to `(DoWhile (Ctx a in) pred out)`
- Constant input substitution
  - `(Ctx (InThen pred (Const c)) Arg)` is equivalent to `(Ctx (InThen pred (Const c)) (Const c))`

*See here for more explanation of the modeled IR/semantics, which are more general (context nodes, arbitrary pure ops) than those actually used in eggcc now:
https://uw-cse.slack.com/archives/C06CH261JM9/p1718175743886959

### DIFF
--- a/proofs/proofs.v
+++ b/proofs/proofs.v
@@ -1,0 +1,240 @@
+(* Coq 8.19.2 *)
+
+Require Import Arith Lia List String.
+
+Inductive Expr : Type :=
+| Arg: Expr
+| Num: nat -> Expr
+| Boolean: bool -> Expr
+| Add: Expr -> Expr -> Expr
+| LessThan: Expr -> Expr -> Expr
+(* Tuples *)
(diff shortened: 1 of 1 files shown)


oflatt (approved):
Yay!

PR #629, opened 2024-10-05
fix linux install script

ryan-berger:


oflatt:
I don't have a linux machine but I assume this works

oflatt (approved):

PR #630, opened 2024-10-07
Upgrade eggcc to deterministic egglog

oflatt:
We now use this egglog PR: https://github.com/egraphs-good/egglog/pull/439


yihozhang (approved):
great!

PR #631, opened 2024-10-15
Add select rules

ajpal:
# `if P A B --> select P A B`

## When A and B are inputs to the if region
Before:
```
# ARGS: 20 30
@main(x: int, y: int) {
  prd: bool = mul x y;
  twenty: int = const 20;
  cmp: bool = lt prd twenty;
  br cmp .then .else;
  .then:
    res: int = id x;
    jmp .done;
  .else:
    res: int = id y;
    jmp .done;
  .done:
    print res;
}
```
After:
```
@main(v0: int, v1: int) {
.b2_:
  v3_: int = mul v0 v1;
  c4_: int = const 20;
  v5_: bool = lt v3_ c4_;
  v6_: int = select v5_ v0 v1;
  print v6_;
}
```

## When A and B are constants
Before:
```
# ARGS: 20 30
@main(x: int, y: int) {
  prd: bool = mul x y;
  twenty: int = const 20;
  cmp: bool = lt prd twenty;
  br cmp .then .else;
  .then:
    res: int = const 4;
    jmp .done;
  .else:
    res: int = const 5;
    jmp .done;
  .done:
    print res;
}
```
After:
```
@main(v0: int, v1: int) {
.b2_:
  v3_: int = mul v0 v1;
  c4_: int = const 20;
  v5_: bool = lt v3_ c4_;
  c6_: int = const 4;
  c7_: int = const 5;
  v8_: int = select v5_ c6_ c7_;
  print v8_;
}
```


dag_in_context/src/optimizations/switch_rewrites.egg
```diff
@@ -29,6 +31,34 @@
       ((union (Get if_e k) (Bop (Smax) a b)))
       :ruleset switch_rewrite) 
 
+; if pred then a else b ~~> (select pred a b)
+; where a and b are inputs to the region
+(rule (
```
oflatt, line 36: I think it's possible for this rule to fork memory if memory is passed through in both branches
To fix this, just make sure that the indices for the arg are not equal

oflatt (changes requested):
Two problems currently:
- linearity issue I mentioned
- CI shows some errors

Otherwise looks great! Love how simple the rules are

oflatt (approved):
Looks good! Would be interested if we get an improvement on anything in nightly

PR #632, opened 2024-10-16
Measure cycles instead of time for benchmarking

oflatt:
This PR does several things (sorry):
- changes the runtime to expose functions that measure cycles on arm and x86
- changes eggcc to use the updated bril fork that instruments the llvm code to use these
- changes all the benchmarks to not have loops that run them multiple times
- changes the benchmarks to have a single print statement before existing to remove I/O overhead
- copies polybench benchmarks into test folder so that we still test the printed matrices
- fixes parallelism in nightly so that it exits on first error


benchmarks/passing/bril/core/euclid.bril
```diff
@@ -1,30 +1,6 @@
-# ARGS: 2550
+# ARGS: 38432 25
```
yihozhang, line 1: 38432 and 25 are coprime so this benchmark is now too easy?

benchmarks/passing/bril/core/fact.bril
```diff
@@ -1,27 +1,10 @@
-# ARGS: 9500
```
yihozhang, line 1: why's the input smaller now?

benchmarks/passing/bril/core/gcd.bril
```diff
@@ -1,36 +1,13 @@
-# ARGS: 1190
+# ARGS: 1190 83
```
yihozhang, line 1: 1190 and 83 are coprimes.

benchmarks/passing/bril/core/euclid.bril
yihozhang: Overmind, coprime actually takes longer for gcd to compute

yihozhang (approved):
lgtm

benchmarks/passing/bril/core/fact.bril
oflatt: It was overflowing or something, but it didn't need to be that big.

PR #636, opened 2024-10-25
Add sequential compiler mode to eggcc and nightly

oflatt:


dag_in_context/src/schedule.rs
```diff
@@ -33,38 +33,88 @@ pub(crate) fn helpers() -> String {
...
+pub fn mk_sequential_schedule() -> Vec<String> {
+    let helpers = helpers();
+    optimizations()
+        .iter()
+        .map(|optimization| {
+            format!(
+                "
+(run-schedule
+   {helpers}
+   {optimization})
```
yihozhang, line 104: also need to run helpers after optimization for `SubsumeIf` / `Subst`?

yihozhang: The full helper would be an overkill actually but some relevant post-processing rulesets seem necessary

oflatt: That's a good point
Currently in nightly sequential does just as well though even without this XD

yihozhang (approved):

PR #639, opened 2024-10-29
Add summary table and compile times to nightly

oflatt:
- Adding compile times ended up causing a refactor of the nightly. It's now a two step process: eggcc, then llvm. This is cleaner in hindsight than using complex run modes
- Visualizations now preserve inputs to enable the change above
- Added an aggregate table, and refactored the old table slightly to share code

### DIFF
--- a/infra/nightly-resources/chart.js
+++ b/infra/nightly-resources/chart.js
@@ -62,6 +62,13 @@ function getEntry(benchmark, runMode) {
   }
 }
 
+function speedup(entry, baseline) {
+  const baseV = mean_cycles(baseline["cycles"]);
+  const expV = mean_cycles(entry["cycles"]);
+  // If you change this, also change the displayed formula in index.html
+  return baseV / expV;
+}
+
 function getValue(entry) {
   if (GLOBAL_DATA.chart.mode === "absolute") {
     return mean_cycles(entry["cycles"]);
@@ -70,10 +77,7 @@ function getValue(entry) {
     if (!baseline) {
       addWarning(`No speedup baseline for ${benchmark}`);
     }
-    const baseV = mean_cycles(baseline["cycles"]);
-    const expV = mean_cycles(entry["cycles"]);
-    // If you change this, also change the displayed formula in index.html
-    return baseV / expV;
+    return speedup(entry, baseline);
   } else {
     throw new Error(`unknown chart mode ${GLOBAL_DATA.chart.mode}`);
   }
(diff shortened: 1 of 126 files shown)


infra/nightly-resources/chart.js
```diff
@@ -16,7 +16,7 @@ const BASELINE_MODE = "llvm-O0";
 
 // Given a list of integers, compute the mean
 // number of cycles
-function mean_cycles(cycles) {
+function mean(cycles) {
```
ajpal, line 19: nit: since this fn isn't actually specific to cycles, can the param/comment etc just talk about mean of a list of ints generally?

infra/profile.py
```diff
@@ -69,16 +72,41 @@ def setup_benchmark(name):
...
+  optimized_bril_file = f'{profile_dir}/{benchmark.treatment}-optimized.bril'
+
+  # get the commands we need to run
+  (eggcc_run_mode, llvm_args) = get_eggcc_options(benchmark)
+  llvm_out_dir = f"{DATA_DIR}/llvm/{benchmark.name}/{benchmark.treatment}"
+
+  cmd1 = f'{EGGCC_BINARY} {benchmark.path} --run-mode {eggcc_run_mode}'
+  cmd2 = f'{EGGCC_BINARY} {optimized_bril_file} --add-timing {llvm_args} -o {profile_dir}/{benchmark.treatment} --llvm-output-dir {llvm_out_dir}'
+
+  print(f'Running: {cmd1}', flush=True)
```
ajpal, line 92: probably delete?

infra/update-frontend.sh
```diff
@@ -0,0 +1,17 @@
+# smalls script to test changes to the javascript or html files
```
ajpal, line 1: What's the purpose of this script?
(As always, I view adding new bash scripts with some skepticism.... 👀)

ajpal (approved):
a few comments, none blocking!
lgtm, but I didn't look super carefully. Lmk if you want a more detailed code review and I can take another look tomorrow

infra/profile.py
oflatt: I find this useful to see what it's really doing, but it's personal preference

infra/update-frontend.sh
oflatt: I don't really love this script but it's super useful for developing the front end. Added a comment- it copies the html and javascript into the output nightly.
Otherwise, you have to run an entire nightly to see changes to the javascript

PR #642, opened 2024-10-30
Fix radio buttons on firefox

ajpal:


chandrakananandi (approved):
Wow you are speedy! Thanks for the fix!! :shipit:

oflatt (approved):
💀 this is scary, how did you know this

ajpal:
Chandra and I had to zoom to figure out the repro since I couldn’t reproduce the issue. That was a big clue that it’d be a browser-specific thing, and there are kind of a lot of Firefox-specific browser things, so that wasn’t a big surprise. Then googling around a little, it seems like Firefox handles partially completed form elements differently than other browsers. The fix is easy- just force the radio button to be how we want it on page load, which will be a no op on chrome since it already does the right thing, and will fix this Firefox-specific issue

PR #644, opened 2024-10-30
Eliminate common subexpressions in our queries

yihozhang:
Porting part of #617 to the main.

Many of our bottleneck queries can indeed be improved because they have common sub-expressions. A better egglog should do this optimization automatically but it currently does not.


oflatt (approved):
Seems good, but yeah I miss the old e-graph based query optimizer

PR #653, opened 2024-11-01
Better select rule

ajpal:
Subsumes the old rules.
`select_simple.rs` before optimization
image

after optimization
image

also `gcd.bril` is now just a big loop with selects, no ifs:
image


oflatt (changes requested):
Looks awesome! Now that it's a more serious optimization, could you give it a separate ruleset?

PR #654, opened 2024-11-06
Add subtuple helper to remove layers of Concat, fix bug in loop inv motion

oflatt:


oflatt:
Looks like there are more bad substitutions to track down!

oflatt:
Fixed another substitution bug, and also fixed the snapshot tests to have "snapshot" in the name for easier testing

dag_in_context/src/optimizations/loop_strength_reduction.egg
```diff
@@ -70,37 +70,38 @@
...
-            (Bop
-                (Mul)
-                (Subst (TmpCtx) (Arg new-arg-ty (TmpCtx)) c-out)
-                (Get (Arg new-arg-ty (TmpCtx)) i)))
+            (Bop (Mul) new-c (Get replace-arg i)))
         (union (Get (Arg new-arg-ty (TmpCtx)) n) new-mul)
 
         ; Subsume the multiplication in the new loop to prevent
         ; from firing loop strength reduction again on the new loop
+        ; subsumption needs something to subsume, add it here
```
ajpal, line 99: in the comment, link to the egglog issue

dag_in_context/src/utility/util.egg
```diff
@@ -68,7 +68,8 @@
 (ruleset subsume-after-helpers)
 ;; After running the `saturating` ruleset, these if statements can be subsumed
 (relation ToSubsumeIf (Expr Expr Expr Expr))
-(rule ((ToSubsumeIf a b c d))
+(rule ((ToSubsumeIf a b c d)
+       (If a b c d))
```
ajpal, line 72: link to the egglog issue here too

ajpal (approved):
two little comments, but lgtm!
thanks for walking through it with me

dag_in_context/src/schedule.rs
```diff
@@ -15,26 +14,33 @@ pub(crate) fn helpers() -> String {
...
+        apply-drop-unions
+        cleanup-drop
 
-    (saturate subst)
-    apply-subst-unions
-    cleanup-subst
-)
+        (saturate subst)
+        apply-subst-unions
+        cleanup-subst)
```
yihozhang, line 30: `subst` generates `AddContext` that needs to be resolved by the `context` ruleset

oflatt: Good catch!

PR #656, opened 2024-11-06
Optimize tail-recursive functions with a base case to loops

oflatt:


tests/snapshots/files__block-diamond-optimize.snap
```diff
@@ -8,22 +8,31 @@ expression: visualization.result
   c2_: int = const 1;
   c3_: int = const 2;
   v4_: bool = lt v0 c3_;
-  c5_: int = const 4;
-  v6_: int = select v4_ c5_ c2_;
-  v7_: int = id v6_;
+  c5_: int = const 0;
```
yihozhang, line 11: this seems to be a small regression

dag_in_context/src/optimizations/swap_if.egg
```diff
@@ -0,0 +1,27 @@
...
+
+
+;; for if statements with two outputs, swaps the order
+;; of the outputs
+(rule
+  ((= lhs (If pred inputs then else))
+   (= (tuple-length then) 2)
+   (= (tuple-length else) 2))
+  (
+    (union
```
yihozhang, line 21: Would it be more useful to union `lhs` with `(Concat (Single (Get new-lhs 1)) (Single (Get new-lhs 0)))` where `new-lhs` is the if expression with output swapped? After all, the expression being unioned here, `(Concat (Single (Get lhs 1)) (Single (Get lhs 0)))`, isn't part of the final program.

dag_in_context/src/schedule.rs
```diff
@@ -91,6 +91,22 @@ pub fn mk_sequential_schedule() -> Vec<String> {
...
+      {helpers}
+      swap-if)
+  {helpers}
+  rec-to-loop
+  {helpers})"
+    ));
+    res.push(format!(
+        "
+;; HACK: when INLINE appears in this string
+;; we perform inlining in this pass
```
yihozhang, line 107: To make this comment clear, "when INLINE appears in this string, we perform inlining in this pass" => "when the keyword INLINE appears in the schedule, `optimize` in `lib.rs` inlines the input program before running this schedule"

Alternatively, we can have this function returns a `Schedule` struct, which makes this no longer a hack.

yihozhang (approved):

dag_in_context/src/optimizations/swap_if.egg
oflatt: I believe that our tuple helpers, which create gets for all of the tuples, do discover that equivalence

dag_in_context/src/schedule.rs
oflatt: Good idea- will follow up with another PR

oflatt:
Merging to unblock you @yihozhang

tests/snapshots/files__block-diamond-optimize.snap
oflatt: will look into it

PR #660, opened 2024-11-11
Separate optimization of functions

oflatt:
This PR adds separate optimization of functions. This allows us to run the egraph separately for different functions, while still doing inlining and all our optimizations we did before.
We could also in the future choose which functions to batch together.

The extractor is now aware of what function it is extracting, which I plan to leverage soon for a better cost model of recursive functions.


dag_in_context/src/lib.rs
```diff
@@ -116,11 +116,15 @@ pub fn print_with_intermediate_vars(termdag: &TermDag, term: Term) -> String {
...
 }
 
-// It is expected that program has context added
+// Build an egglog program that optimizes a particular batch of functions `fns`
+// with a schedule `schedule`.
+// If `inline_program` is true, it also inlines calls in `fns`.
+// `inline_program` is the program to inline calls from, allowing us to inline unoptimized function bodies.
 pub fn build_program(
     program: &TreeProgram,
+    inline_program: Option<&TreeProgram>,
```
yihozhang, line 125: Do we need both program and inline_program? How are they different? Is `inline_program` the "full" program and `program` a batch of programs to be optimized?

tests/snapshots/files__gamma_condition_and-optimize.snap
```diff
@@ -14,21 +14,13 @@ expression: visualization.result
   br v3_ .b8_ .b9_;
 .b8_:
   c10_: int = const 3;
-  v11_: int = id c10_;
-  br v4_ .b12_ .b13_;
-.b12_:
-  v11_: int = id c5_;
+  v11_: int = select v4_ c5_ c10_;
```
yihozhang, line 17: Nice that the `select` optimization is applied. Any idea why it was not applied before?

oflatt: It's flaky, probably an extraction problem or same cost

dag_in_context/src/lib.rs
oflatt: The inline program is the original program- if you make them the same, you might inline something that has already been inlined, causing blowup

PR #661, opened 2024-11-12
Separate llvm compilation into optimization and lowering

oflatt:
This PR allows eggcc to run LLVM with separate optimization levels for the middle-end and lowering to a binary. The idea is to give us a more fair comparison in the nightly because LLVM won't be doing platform-specific optimization, so we compare more directly to the llvm middle-end optimizer. 

It leaves in the normal O3 run as `llvm-O3-O3`.

### DIFF
--- a/benchmarks/passing/bril/long/function_call.bril
+++ b/benchmarks/passing/bril/long/function_call.bril
@@ -1,8 +1,5 @@
 # ARGS: 25
 
-# failing due to brillvm bug:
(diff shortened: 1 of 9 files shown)


ajpal (approved):
lgtm but probably worth waiting for a nightly to make sure the frontend still works.

oflatt:
Nightly is dedicated to PLDI push right now so it's tied up. I'll keep an eye on the nightly for main

PR #662, opened 2024-11-13
Go back to batching all fns from the program together

oflatt:
Pr #660 separated the egraph's optimization of different functions. Turns out this makes everything a lot slower, perhaps due to sharing of types and other subexpressions.
This PR changes it back to batching all the functions together, but keeps extraction as per-function. This allows the extractor in the future to be function-aware, making recursive calls expensive.

It also optimizes extraction to find reachable regions instead of extracting everything in the egraph.


ajpal (approved):
lgtm, but I'm not that familiar with this part of the code....

PR #664, opened 2024-11-15
Move the passthrough rules out of helpers, simplifying the schedule

oflatt:


tests/snapshots/files__if_dead_code_nested-optimize.snap
```diff
@@ -12,49 +12,89 @@ expression: visualization.result
   br v3_ .b6_ .b7_;
 .b6_:
   v8_: bool = lt v0 c2_;
-  c9_: int = const 1;
```
yihozhang, line 15: Is this a regression? Code seems longer

oflatt: good point, I'll investigate

oflatt: it seems be just be flaky, I think anjali will fix it with the next select pr

yihozhang (approved):

PR #666, opened 2024-11-18
Simplify schedule by separating out the saturating combined ruleset

oflatt:
This PR is a stepping stone for making our schedule faster. It simplifies things by listing out the rulesets explicitly, and is a good intermediate point for #663


yihozhang (approved):
Great!

PR #668, opened 2024-11-19
Remove duplicates in substitution rules

yihozhang:


oflatt (approved):
LGTM
We should consider optimizing these on the egglog side though

PR #670, opened 2024-11-20
Sequentially apply loop inversion

yihozhang:
Interestingly, this seems to undo some of the smax and select optimizations, as well as a loop strength reduction test

edit: there's also a big bump on CI time, probably because loop inversion enables a bunch of other optimizations (which is not extracted nonetheless)

### DIFF
--- a/dag_in_context/src/schedule.rs
+++ b/dag_in_context/src/schedule.rs
@@ -83,25 +83,39 @@ pub fn rulesets() -> String {
 pub fn mk_sequential_schedule() -> Vec<String> {
     let helpers = helpers();
 
-    let mut res = vec![
-        format!(
-            "
+    let mut res = vec![format!(
+        "
 (run-schedule
    (saturate
       {helpers}
       passthrough
       state-edge-passthrough))"
-        ),
-        format!(
-            "
+    )];
+    res.push(format!(
+        "
 (run-schedule
-    (repeat 2
+  (repeat 2
+    {helpers}
+    loop-inversion)
+  
+  {helpers})"
+    ));
+    res.push(format!(
+        "
+(run-schedule
+  (repeat 2
       {helpers}
-      loop-inversion
-      {helpers}
-    ))"
-        ),
-    ];
+      swap-if)
+  {helpers}
+  rec-to-loop
+  {helpers})"
+    ));
+    res.push(format!(
+        "
+;; HACK: when INLINE appears in this string
+;; we perform inlining in this pass
+(run-schedule {helpers})"
+    ));
     res.extend(optimizations().iter().map(|optimization| {
         format!(
             "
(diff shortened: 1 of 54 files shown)


tests/snapshots/files__loop_with_mul_by_inv-optimize.snap
```diff
@@ -11,20 +11,18 @@ expression: visualization.result
   v5_: int = id c2_;
   v6_: int = id c3_;
   v7_: int = id v0;
-  v8_: int = id c2_;
-.b9_:
-  v10_: int = add v4_ v8_;
+.b8_:
+  v9_: int = mul v5_ v6_;
```
yihozhang, line 15: This `mul` used to be optimized away by LSR

dag_in_context/src/schedule.rs
```diff
@@ -104,19 +115,29 @@ pub fn mk_sequential_schedule() -> Vec<String> {
...
-/// a schedule that runs optimizations over the egraph.
 pub fn parallel_schedule() -> Vec<String> {
     let helpers = helpers();
 
-    vec![format!(
-        "
+    vec![
+        format!(
+            "
 (run-schedule
```
yihozhang, line 124: This is currently doing
```
vec![
passthrough,
loop-inversion,
main-rulesets
]
```
After changing to 
```
vec![
loop-inversion,
passthrough,
main-rulesets
]
```
It does not terminate, interesetingly

oflatt: What benchmark doesn't terminate? It might be related to dead code, the thing I mentioned last meeting

oflatt:
Probably the big bump in CI time is due to there being two egraph passes instead of one, right? It's the time spent running all the helpers
Hopefully we can reduce with some scheduling

oflatt:
We probably should figure out why the `smin` optimization isn't firing any more, since there are not any loops in that example right?

src/util.rs
```diff
@@ -808,14 +808,17 @@ impl Run {
...
+                }
 
-                let egglog =
-                    build_program(&dag, Some(&dag), &dag.fns(), &mut cache, &schedule_steps[0]);
+                let egglog = build_program(
+                    &dag,
+                    Some(&dag),
+                    &dag.fns(),
+                    &mut cache,
+                    &schedule_steps.join("\n"),
```
oflatt, line 820: I think this is a bad idea since it's not something eggcc does. Probably we should make this give you the last egglog file, or it should work with `stop-after-n-passes`

oflatt (approved):
Approved! But could you please fix the comment I had as a follow-up?

src/util.rs
yihozhang: Will address this in my next PR

PR #671, opened 2024-11-21
Little updates to the chart

ajpal:
- Make the yellow bar actually yellow
- Make the error bars a little thinner and lighter
- We don't have looped benchmarks anymore, so get rid of those buttons
- It's nice to be able to look at the fastest or slowest benchmarks, so add easy filters for that


oflatt (approved):
looks good!

PR #672, opened 2024-11-21
Simplify helpers

yihozhang:
Part of #663 

This PR separates `drop` and `subst` out of the big saturating loop of `always-run` by taking `is-resolved` out of `always-run`. It also breaks down the big saturation loop associated with always-run, canon, interval-analysis, always-switch-rewrite, loop-iters-analysis, etc.

### DIFF
--- a/dag_in_context/src/schedule.rs
+++ b/dag_in_context/src/schedule.rs
@@ -2,21 +2,6 @@ pub(crate) fn helpers() -> String {
     "
     ;; first, run substitution and drop to saturation
     ;; these depend on type analysis, always-run, and context
-    ;; (saturate
-    ;;     (saturate
-    ;;       type-analysis
-    ;;       (saturate type-helpers)
-    ;;       error-checking
-    ;;       always-run
-    ;;       context)
-        
-    ;;     (saturate drop)
-    ;;     apply-drop-unions
-    ;;     cleanup-drop
-
-    ;;     (saturate subst)
-    ;;     apply-subst-unions
-    ;;     cleanup-subst)
 
     ;; first, saturate always run
     (saturate
--- a/dag_in_context/src/schedule.rs
+++ b/dag_in_context/src/schedule.rs
@@ -2,41 +2,60 @@ pub(crate) fn helpers() -> String {
     "
     ;; first, run substitution and drop to saturation
     ;; these depend on type analysis, always-run, and context
-    (saturate
-        (saturate
-          type-analysis
-          (saturate type-helpers)
-          error-checking
-          always-run
-          context)
+    ;; (saturate
+    ;;     (saturate
+    ;;       type-analysis
+    ;;       (saturate type-helpers)
+    ;;       error-checking
+    ;;       always-run
+    ;;       context)
         
-        (saturate drop)
-        apply-drop-unions
-        cleanup-drop
+    ;;     (saturate drop)
+    ;;     apply-drop-unions
+    ;;     cleanup-drop
+
+    ;;     (saturate subst)
+    ;;     apply-subst-unions
+    ;;     cleanup-subst)
+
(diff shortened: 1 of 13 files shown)


yihozhang:
`optimizations::loop_unroll::loop_unroll_simple` in `dag_in_context` is failing

dag_in_context/src/schedule.rs
```diff
@@ -45,10 +51,14 @@ fn cheap_optimizations() -> Vec<String> {
     // TODO enable loop peeling
     // currently causes saturation issues, probably by creating dead loops that are allowed to have any value
 
-    ["loop-simplify", "memory", "peepholes"]
-        .iter()
-        .map(|opt| opt.to_string())
-        .collect()
+    [
+        "loop-simplify",
+        // "memory",
```
yihozhang, line 56: since we are not running the memory helpers anyway

tests/snapshots/files__loop_with_mul_by_inv-optimize-sequential.snap
```diff
@@ -11,20 +11,18 @@ expression: visualization.result
   v5_: int = id c2_;
   v6_: int = id c3_;
   v7_: int = id v0;
-  v8_: int = id c2_;
-.b9_:
-  v10_: int = add v4_ v8_;
+.b8_:
+  v9_: int = mul v5_ v6_;
```
oflatt, line 15: This seems like a regression? There used to be no mul in the optimized one afaict
Not sure which optimization does this transformation

dag_in_context/src/schedule.rs
```diff
@@ -2,41 +2,47 @@ pub(crate) fn helpers() -> String {
...
-        always-switch-rewrite
-        loop-iters-analysis
-        ; memory-helpers TODO run memory helpers for memory optimizations
+        (saturate drop)
+        apply-drop-unions
+        cleanup-drop
     )
 
+    (saturate canon)
+    (saturate interval-analysis)
```
oflatt, line 35: do we need to do types again here? Idk if interval analysis depends on canon

oflatt (changes requested):
Small worry about a snapshot test, otherwise looks great

oflatt:
Will be curious if this helps ci time

dag_in_context/src/schedule.rs
yihozhang: `canon` is just doing some equational rewriting, so I think this just means we are missing some type and interval information for the new subexpressions generated in the current round, which probably won't matter I think

tests/snapshots/files__loop_with_mul_by_inv-optimize-sequential.snap
yihozhang: Oh interesting... I run `cargo run --release tests/passing/small/loop_with_mul_by_inv.bril` locally and it generates the old instructions which does not have `mul`. Must be nondeterminism + greediness of our extractor.


Bril code generated locally:
```
# ARGS: 4
@main(v0: int) {
.b1_:
  c2_: int = const 0;
  c3_: int = const 5;
  v4_: int = id c2_;
  v5_: int = id c2_;
  v6_: int = id c3_;
  v7_: int = id v0;
  v8_: int = id c2_;
.b9_:
  v10_: int = add v4_ v8_;
  c11_: int = const 1;
  v12_: int = add c11_ v5_;
  v13_: int = add v6_ v8_;
  v14_: bool = lt v5_ v7_;
  v4_: int = id v10_;
  v5_: int = id v12_;
  v6_: int = id v6_;
  v7_: int = id v7_;
  v8_: int = id v13_;
  br v14_ .b9_ .b15_;
.b15_:
  print v4_;
  ret;
}
```

oflatt (approved):
Looks good!

PR #675, opened 2024-11-21
Print the egglog program before the last schedule for multi-pass schedule in `--run-mode egglog`

yihozhang:
Follow up of #670


dag_in_context/src/lib.rs
```diff
@@ -256,7 +256,10 @@ pub enum Schedule {
 #[derive(Clone, Debug)]
 pub struct EggccConfig {
     pub schedule: Schedule,
-    pub stop_after_n_passes: usize,
+    /// Stop after this many passes.
+    /// If stop_after_n_passes is negative,
+    /// run [0 ... schedule.len() + stop_after_n_passes + 1] passes.
```
oflatt, line 261: This seems weird... I would expect `-1` to run all but the last pass, but it looks like it runs all of them.
Perhaps you should make it i64::MAX by default and use `[0 .. scheduler.len() + stop_after_n_passes ]` in the neg case?

dag_in_context/src/schedule.rs
```diff
@@ -173,8 +173,6 @@ pub fn parallel_schedule() -> Vec<CompilerPass> {
         )),
         CompilerPass::InlineWithSchedule(format!(
             "
-;; HACK: when INLINE appears in this string
```
oflatt, line 176: good catch

src/util.rs
```diff
@@ -802,27 +802,31 @@ impl Run {
...
-                assert_eq!(self.eggcc_config.schedule, Schedule::Parallel, "Output egglog only works in parallel mode. Sequential mode does not use a single egraph");
-
                 let rvsdg = Optimizer::program_to_rvsdg(&self.prog_with_args.program)?;
                 let (dag, mut cache) = rvsdg.to_dag_encoding(true);
 
-                let schedule_steps = parallel_schedule();
-                if schedule_steps.len() != 1 {
-                    log::warn!("Parallel schedule had multiple steps! You may need to adjust the schedule to make eggcc tractable.");
-                }
+                let eggcc_config = EggccConfig {
```
oflatt, line 808: It would be nice if this took into account stop_after_n_passes so that we can debug particular eggcc passes
Right now it just takes the last pass

src/util.rs
```diff
@@ -802,27 +802,31 @@ impl Run {
...
+                    // stop before the last pass.
+                    stop_after_n_passes: -2,
+                    .. self.eggcc_config.clone()
+                };
+                let optimized = dag_in_context::optimize(&dag, &mut cache, &eggcc_config)
```
oflatt, line 813: You could refactor the inner loop of optimize into a function so you can get the egglog file directly instead of calling build_program

oflatt (changes requested):
Some suggested improvments

src/util.rs
yihozhang: Yeah realized this right after I pushed

dag_in_context/src/lib.rs
yihozhang: I updated the code to reflect this, although it is a little less elegant now to handle the pass before `i64::MAX`

src/util.rs
yihozhang: hmm, I'm not sure how to refactor here. Maybe leave it to a future PR?

oflatt: sure

src/util.rs
```diff
@@ -802,27 +802,36 @@ impl Run {
...
+                let eggcc_config = EggccConfig {
+                    // stop before the last pass that user specified.
+                    stop_after_n_passes: stop_after_n_passes - 1,
+                    ..self.eggcc_config.clone()
+                };
+                let optimized = dag_in_context::optimize(&dag, &mut cache, &eggcc_config)
+                    .map_err(EggCCError::EggLog)?;
+
+                let schedules = parallel_schedule();
+                let last_schedule_step = schedules.last().unwrap();
```
oflatt, line 822: You're still using the wrong pass here- didn't you want the stop_after_n_passes pass?

oflatt (changes requested):

src/util.rs
yihozhang: good catch, fixed!

src/util.rs
```diff
@@ -802,27 +802,36 @@ impl Run {
...
-                let schedule_steps = parallel_schedule();
-                if schedule_steps.len() != 1 {
-                    log::warn!("Parallel schedule had multiple steps! You may need to adjust the schedule to make eggcc tractable.");
-                }
+                // how many actual passes to run
+                let cutoff = self
+                    .eggcc_config
+                    .get_normalized_cutoff(parallel_schedule().len());
+                assert!(cutoff != 0);
+                let cutoff = cutoff - 1;
```
oflatt, line 813: Why are you subtracting 1 here? Shouldn't you pass it to the config as-is?

oflatt (changes requested):

src/util.rs
yihozhang: say there are N passes and the user passes `stop-after-n-passes N` (the default value). This will run 0..(N-1) passes, and use the last schedule, which is at position `N-1`. Maybe I miss something here?

oflatt: Ah okay, I get it now

oflatt (approved):
Great work! I just ask that you test this PR before you merge, since the logic is surprisingly subtle
Perhaps we should add a small check that running the result of the egglog run mode is the same as running n passes?

yihozhang:
Added a test

PR #678, opened 2024-11-22
Optimize initial egraph building, which used to take quadratic time

yihozhang:
Right now, the initial database is built with tens of thousands of let statements. Each `let` statement creates a new relation and requires egglog to go through every relation in the database, so it takes quadratic time to rebuild.

This PR wraps all the initialization in a single rule, so no rebuilding is done in-between the lets. ~This also mitigates the complaint @oflatt had earlier about excessive logging.~ (The log is still pretty big because we now print the giant initialization rule as part of the log...)


oflatt (approved):
amazing

PR #680, opened 2024-11-22
Fix schedule order

yihozhang:
interval-analysis and always-switch use `subst`


dag_in_context/src/optimizations/ivt.egg
```diff
@@ -123,6 +123,6 @@
     (union new-expr loop)
     (union new-loop-context (InLoop (Arg argI new-if-true-ctx) new-loop-outputs))
 
-    (delete (DoWhile inpW outW))
```
yihozhang, line 126: to be consistent with other optimizations

oflatt:
Closing in favor of fixing interval analysis to not use subst

PR #681, opened 2024-11-22
Fix schedule order via interval-rewrite rules

yihozhang:
Alternative to #680


dag_in_context/src/optimizations/ivt.egg
```diff
@@ -123,6 +123,6 @@
     (union new-expr loop)
     (union new-loop-context (InLoop (Arg argI new-if-true-ctx) new-loop-outputs))
 
-    (delete (DoWhile inpW outW))
+    (subsume (DoWhile inpW outW))
```
oflatt, line 126: Is this an unrelated change?
I had a weird bug with this this morning that we should investigate where nothing could be extracted

oflatt (changes requested):
Yay!

dag_in_context/src/optimizations/ivt.egg
yihozhang: I also had this in my previous PR. It just makes our rules more consistent (always use subsume rather than delete). 

Your bug might be caused by the wrong schedule order that the other changes in this PR fix? I had a similar issue and that was actually the original motivation for this PR.

oflatt (approved):
Looks good!

PR #682, opened 2024-11-23
Smarter helper rules

yihozhang:
Eggcc has been constantly blowing up with the giant analyses our helpers generate. But do our helpers really need to generate that many facts?

### DIFF
--- a/dag_in_context/src/type_analysis.egg
+++ b/dag_in_context/src/type_analysis.egg
@@ -5,20 +5,34 @@
 (rewrite (TLConcat (TNil) r) r :ruleset type-helpers)
 (rewrite (TLConcat (TCons hd tl) r)
          (TCons hd (TLConcat tl r))
+         :subsume
          :ruleset type-helpers)
 
 (function TypeList-length (TypeList) i64 :unextractable)
 (function TypeList-ith (TypeList i64) BaseType :unextractable)
-(function TypeList-suffix (TypeList i64) TypeList :unextractable)
+; (function TypeList-suffix (TypeList i64) TypeList :unextractable)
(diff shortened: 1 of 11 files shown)


dag_in_context/src/type_analysis.egg
```diff
@@ -5,20 +5,34 @@
 (rewrite (TLConcat (TNil) r) r :ruleset type-helpers)
 (rewrite (TLConcat (TCons hd tl) r)
          (TCons hd (TLConcat tl r))
+         :subsume
          :ruleset type-helpers)
 
 (function TypeList-length (TypeList) i64 :unextractable)
 (function TypeList-ith (TypeList i64) BaseType :unextractable)
-(function TypeList-suffix (TypeList i64) TypeList :unextractable)
```
yihozhang, line 12: The old TypeList-suffix based implementation is quadratic in the size of the list (which easily goes to 20+)

dag_in_context/src/utility/canonicalize.egg
```diff
@@ -56,36 +59,29 @@
...
+         :subsume
          :when ((> b 1))
          :ruleset always-run)
 
-; Also figure out what existing expressions are subtuples of other things
-; this helps remove concat layers
-(rule ((Get expr i))
-      ((union (Single (Get expr i))
-              (SubTuple expr i 1)))
+; ;; This rule makes sure every tuple can be
```
yihozhang, line 77: this is actually unnecessary because we are already generating gets for every "important" tuple. Removed in my new PRs.

oflatt:
closing due to being out of date

PR #684, opened 2024-11-26
Add debugging infrastructure using DebugExpr relation

oflatt:


dag_in_context/src/schema.egg
```diff
@@ -240,3 +240,8 @@
 (bop->string (PtrAdd) "PtrAdd")
 (bop->string (Print) "Print")
 (bop->string (Free) "Free")
+
+
+;; If anything is put in the DebugExpr relation, we'll extract them instead of the original program.
```
ajpal, line 245: Is there a reason you want to do it instead of the original program not in addition?

ajpal (approved):

dag_in_context/src/schema.egg
oflatt: Actually you do want it in addition so you can resolve calls- PR for that coming soon

PR #685, opened 2024-11-27
Fix the check-primes benchmark.

ezrosent:
This change also allows for tests to run on a newer MacOS install.


oflatt (approved):
Awesome! This fix makes sense. We could also consider writing a checker to disallow the case where something is unbound. However I don't actually know how to write this checker and have it be efficient.

PR #687, opened 2024-11-28
Fix dependency

yihozhang:
I am building eggcc in a fresh Ubuntu virtual machine and got the following errors

```
/usr/bin/ld: cannot find -lz
/usr/bin/ld: cannot find -lstdc++
```

The second one is fixed because I don't have a C++ compiler installed, so I immediately fixed it. I found the solution to the first error with [this StackOverflow post](https://stackoverflow.com/questions/3373995/usr-bin-ld-cannot-find-lz).


oflatt:
nice fix!

oflatt (approved):

PR #690, opened 2024-11-30
No quadratic Get population

yihozhang:


dag_in_context/src/utility/util.egg
```diff
@@ -25,35 +25,67 @@
        (= len (TypeList-length tl)))
       ((set (tuple-length expr) len)) :ruleset always-run)
 
+
+;; Leading expressions are tuples that are used as a whole
+;; during optimization and are matched in the query.
+(relation leading-Expr (Expr))
+(relation leading-Expr-list (ListExpr))
```
oflatt, line 32: Is leading-expr-list even used anywhere? perhaps we should cut it?

tests/snapshots/files__implicit-return-optimize.snap
```diff
@@ -53,31 +54,32 @@ expression: visualization.result
   v6_: int = id c1_;
   v7_: int = id c3_;
 .b8_:
-  c9_: int = const 14;
```
oflatt, line 56: huh, I wonder why the old one has const 14 while the new one doesn't

oflatt (changes requested):

tests/snapshots/files__implicit-return-optimize.snap
yihozhang: seems like we missed the constant folding and computed c9_ as 15 - 1 (v10_ in the new program)

dag_in_context/src/utility/util.egg
oflatt: oh great, thx

tests/snapshots/files__implicit-return-optimize.snap
oflatt: That does seem like a regression- do you know why it isn't happening now? Perhaps we actually did need `Get` for a generated concat?
Could be that we need to demand `Get` for all of the generated `Concat`, just not subtuples

yihozhang:
I investigated a bit and I have no idea why the optimization is not firing... Some clues:

This is the program in main vs in this branch
image

image

At first I thought it is the loop invariant code motion. Both branches do the LICM over the `pow` function, which eliminates the `Sub` node within `pow` (but not `main`). For `main`, it turns out that if you just do inlining before it hits the first schedule (either with manual inlining or `InlineWithSchedule`, the passthrough optimization alone (maybe together with some other optimizations) will produce the desired program.

It is a bit weird to me why passthrough can do this optimization, because all it does is unioning the output of the loop with the input of the loop, without giving the body any more information. Maybe there's a magic rule in the helpers that passthrough enables that does this... Need to investigate more.

oflatt:
It's it passthrough combined with the interval analysis?

yihozhang:
I also suspected this, but interval analysis does not match on `DoWhile`s, so I can't think of how the combination of the two does it.

oflatt:
Wow, interesting
I think I'm okay with merging this regression, or you can improve interval analysis to be aware of passthrough values in loops

tests/snapshots/files__if_dead_code_nested-optimize-sequential.snap
```diff
@@ -34,10 +34,30 @@ expression: visualization.result
 .b22_:
   v24_: bool = lt c5_ v0;
   c25_: int = const 4;
-  v26_: int = select v24_ c5_ c25_;
```
yihozhang, line 37: just pulled from main again, this seems like a regression

tests/snapshots/files__implicit-return-optimize.snap
yihozhang: oh nice this is gone after your new IVT is merged

tests/snapshots/files__if_dead_code_nested-optimize-sequential.snap
yihozhang: nvm, this is sequential

oflatt (approved):
Looks good!

PR #691, opened 2024-12-02
Permutations using substituion

oflatt:
This PR refactors IVT to use substitution instead of permutations. 
This hopefully will make it easier to generalize, since this code is cleaner and substitution is more general.


oflatt:
Subsumed by #692

PR #692, opened 2024-12-03
Generalize IVT to loops with passed-through values

oflatt:
This PR generalizes IVT to allow it to handle values that flow through the loop outside the if condition.
This involves carefully moving these values out and handling them separately


dag_in_context/src/utility/canonicalize.egg
```diff
@@ -68,6 +68,22 @@
          :when ((> b 1))
          :ruleset always-run)
 
+; Some of our rules (like ivt.egg) match on `Concat`.
```
yihozhang, line 71: I can make this a default behavior of `leading-Expr`s in my PR

yihozhang (approved):
Nice! A lot of the snapshots are indeed simplified!

PR #697, opened 2024-12-11
Fix bug where invalid context caches are used

oflatt:


yihozhang (approved):

yihozhang:
from Oliver explaining to me: the bug is that we were using caches from the last pass' program, since we are doing multi-pass schedules now. As a result, it does some unexpected unions

PR #698, opened 2024-12-12
Fix bug in IVT, add soundness check

oflatt:
This PR fixes two bugs in IVT:
- IVT was matching on different If statements than the analysis results were for
- IVT was using the wrong number for passthrough arguments in the analysis

It also adds a pretty weak soundness check that is useful nonetheless, checking if top-level constants are equal.


dag_in_context/src/optimizations/ivt.egg
```diff
@@ -42,78 +43,82 @@
...
 
 (rule (
     (= loop (DoWhile inpW outW))
-    (= (IVTNewInputsAnalysis outW if) (IVTAnalysisRes perm pperm passthrough-tys _len))
+    (= (IVTNewInputsAnalysis outW ifnode) (IVTAnalysisRes perm pperm passthrough-tys _len))
+    (= ifnode (IfNode if if-cond if-inputs then else))
     (= if-inputs-len (tuple-length if-inputs))
     (= passthrough-len (TypeList-length passthrough-tys))
-    (= if (If if-cond if-inputs then else))
+    (= if-len (tuple-length if))
```
yihozhang, line 121: if-len is not used in the rule?

yihozhang (approved):
The two changes make sense to me. I think the code for IVT is getting more and more complex that it may be hard to maintain in the future...

Maybe if one day we can make the combination of passthrough + IVT scales well, then we don't need this version of IVT? Or do you think even if that's the case, this version of IVT is also superior in some ways. Guess once we make eggcc scalable enough, we can just run our benchmark on both versions of the IVT and see.

Great work!

dag_in_context/src/optimizations/ivt.egg
oflatt: good catch

oflatt:
how would it work without this generalization? I think there will always be passthrough values hanging around that are used in the loop

yihozhang:
Doesn't running IVT before passthrough solve the issue?

PR #700, opened 2025-01-07
make jitter plots collapsible

ajpal:


oflatt (approved):
ok but button is evil

ajpal:
> ok but button is evil

Why? I legitimately don't really see a problem with this way of doing it lol

PR #702, opened 2025-01-09
Add `abs` and make Oliver's example work

ajpal:
```
// ARGS: 20
fn unrelated_fn(input: i64, res: i64) -> i64 {
    return input / (res / 3);
}

fn other_unrelated_fn(input: i64, res: i64) -> i64 {
    return res / (input / 5);
}

fn main(input: i64) {
    let mut res: i64 = abs(input) * 2;

    if (input > 0) {
        res = res + abs(input) - input;
        res = unrelated_fn(input, res);
    } else {
        res = res + abs(input) + input;
        res = other_unrelated_fn(input, res);
    }

    println!("{}", res);
}
```

Output bril:
```
# ARGS: 20
@unrelated_fn(v0: int, v1: int): int {
.b2_:
  c3_: int = const 3;
  v4_: int = div v1 c3_;
  v5_: int = div v0 v4_;
  ret v5_;
}
@other_unrelated_fn(v0: int, v1: int): int {
.b2_:
  c3_: int = const 5;
  v4_: int = div v0 c3_;
  v5_: int = div v1 v4_;
  ret v5_;
}
@main(v0: int) {
.b1_:
  c2_: int = const 0;
  v3_: bool = gt v0 c2_;
  v4_: int = abs v0;
  c5_: int = const 2;
  v6_: int = mul c5_ v4_;
  br v3_ .b7_ .b8_;
.b7_:
  c9_: int = const 3;
  v10_: int = div v6_ c9_;
  v11_: int = div v0 v10_;
  v12_: int = id v11_;
  print v12_;
  ret;
  jmp .b13_;
.b8_:
  c14_: int = const 5;
  v15_: int = div v0 c14_;
  v16_: int = div v6_ v15_;
  v12_: int = id v16_;
  print v12_;
  ret;
.b13_:
}
```

SVGs
Before
image

After
image

### DIFF
--- a/dag_in_context/src/interval_analysis.egg
+++ b/dag_in_context/src/interval_analysis.egg
@@ -168,10 +168,10 @@
         (= lhs (Bop (Mul) x y))
         (= (IntB hi-x) (hi-bound x))
         (= (IntB hi-y) (hi-bound y))
-        (<= hi-x 0)
-        (<= hi-y 0) 
+        (< hi-x 0)
+        (< hi-y 0) 
       )
-      ((set (lo-bound lhs) (IntB 0)))
+      ((set (lo-bound lhs) (IntB 1)))
       :ruleset interval-analysis)
 
 ; negative * positive is negative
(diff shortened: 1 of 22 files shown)


dag_in_context/src/interval_analysis.egg
```diff
@@ -214,6 +247,35 @@
       ((set (hi-bound lhs) (BoolB (bool-< la hb))))
       :ruleset interval-analysis)
 
+; Abs
+
+; absolute value is non-negative
```
oflatt, line 252: it's not true 😭

oflatt (changes requested):
Good but one of your rules seems unsound

oflatt (approved):
Looks good, you just have to fix CI
Look over the snapshots for our example carefully!

PR #703, opened 2025-01-15
Fixes the cost calculation which counts all `Mul` in a region only once

yihozhang:
This is the loop strength reduction example after this PR

image

The cost of this is now computed as 27026, which is roughly 300 * (50 + 10 + 10 + 10 + 10) (the cost of add and lt is 10, and of print is 50)

Unfortunately, in this example, only LSR is applied, but not loop unrolling (which used to apply), because we only run our optimization twice. We can either bump the #iter opt is applied or the #batch to have loop unrolling and LSR both applied


oflatt (approved):
Clever solution! Improves nightly very slightly! Great work

PR #707, opened 2025-01-22
Update bitwise-ops

yihozhang:
so that it runs `xor` multiple times.

### DIFF
--- a/benchmarks/passing/bril/core/bitwise-ops.bril
+++ b/benchmarks/passing/bril/core/bitwise-ops.bril
@@ -71,26 +71,12 @@
   one: int = const 1;
   zero: int = const 0; 
 
-  sel: int = sub c one;
-
-  ans: int = id zero; 
-
-  less: bool = lt sel zero;
-  equal: bool = eq sel zero;
-  greater: bool = gt sel zero; 
-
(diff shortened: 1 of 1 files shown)


benchmarks/passing/bril/core/bitwise-ops.bril
```diff
@@ -71,26 +71,12 @@
...
-  less: bool = lt sel zero;
-  equal: bool = eq sel zero;
-  greater: bool = gt sel zero; 
-
-  br less .and_op .useless_lbl;
-.useless_lbl:
-  br equal .or_op .xor_op;
-
-.and_op:
-  ans: int = call @AND a b;
```
oflatt, line 87: we don't call AND and OR anymore? shouldn't we keep the benchmark closer to the original, just add a loop around it?

oflatt (changes requested):

benchmarks/passing/bril/core/bitwise-ops.bril
yihozhang: We didn't directly call AND and OR before as well (because of the fixed input). XOR is implemented in terms of AND and OR

oflatt (approved):
thanks!

PR #708, opened 2025-01-23
Latex macro for percent of benchmarks where we recover 80% of the performance improvement that llvm finds

oflatt:
image


ajpal (approved):
I'm a little skeptical that this is a useful thing to report on, but fine to merge if you want it

PR #709, opened 2025-01-23
Conditional Invariant Code Motion (Pull Out)

ajpal:
With this rule:
```
(rule ((Bop (Mul) (Uop (Abs) x) (Const (Int 3) ty ctx)))
      ((panic "good")) :ruleset peepholes)
```
we can tell that the optimization is firing and finding the way to pull `res = res + abs(input)` out of both branches, but due to extraction issues, we're not extracting that program.

### DIFF
--- a/tests/snapshots/files__if_invariant_dont_pull_out-optimize-sequential.snap
+++ b/tests/snapshots/files__if_invariant_dont_pull_out-optimize-sequential.snap
@@ -24,26 +24,25 @@ expression: visualization.result
   v4_: int = abs v0;
   c5_: int = const 2;
   v6_: int = mul c5_ v4_;
-  br v3_ .b7_ .b8_;
-.b7_:
-  v9_: int = sub v6_ v0;
-  v10_: int = add v0 v9_;
-  c11_: int = const 3;
-  v12_: int = div v10_ c11_;
-  v13_: int = div v0 v12_;
-  v14_: int = id v13_;
-  print v14_;
+  c7_: int = const 3;
+  v8_: int = mul c7_ v4_;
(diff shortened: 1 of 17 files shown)


oflatt:
Are you able to write a small snapshot test where we do end up extracting the result?

ajpal:
New snapshot test that shows conditional invariant code motion in action:
```
// ARGS: 20
fn main(x: i64) {
    let a: i64 = x * 5;
    if (x > 0) {
        let r: i64 = x * 5;
        let z: i64 = 20;
        println!("{}", z);
    } else {
        let r: i64 = x * 5;
        let z: i64 = 20;
        println!("{}", z);
    }
    println!("{}", r);
}

```

Before
image

After
image

We've added a new input to the region (`x * 5`)
Note that the now-unused input `x` is still there, because we can't detect that it's now unused

oflatt (approved):
amazing

tests/snapshots/files__if_invariant_dont_pull_out-optimize-sequential.snap
```diff
@@ -24,26 +24,25 @@ expression: visualization.result
...
-  br v3_ .b7_ .b8_;
-.b7_:
-  v9_: int = sub v6_ v0;
-  v10_: int = add v0 v9_;
-  c11_: int = const 3;
-  v12_: int = div v10_ c11_;
-  v13_: int = div v0 v12_;
-  v14_: int = id v13_;
-  print v14_;
+  c7_: int = const 3;
```
oflatt, line 27: looks like this broke 😭

oflatt (approved):
we can merge for now but we need to fix the cost model

PR #711, opened 2025-01-26
Remove empty entry blocks

ezrosent:
This is a common feature of benchmarks that introduce translation overhead.


ezrosent:
The big change here is the one in `to_bril`

It changed >150 snapshots, which is why the diff is so large. This appears to really improve translation overhead for some benchmarks.

oflatt (approved):
great!

PR #713, opened 2025-01-28
Conditional Push In

ajpal:
image

Although, we don't remove the original input, so it ends up actually looking like this:
image

This extra (now unused) input will hopefully get cleaned up with Oliver's new dead-code-aware extraction.

### DIFF
--- a/dag_in_context/src/optimizations/conditional_push_in.egg
+++ b/dag_in_context/src/optimizations/conditional_push_in.egg
@@ -1,29 +1,16 @@
 (ruleset push-in)
 
-; push constant input into region
-(rule (
-        (= if_e (If pred inputs thn els))
-        (= (Const c ty ctx) (Get inputs i))
-        (HasArgType thn arg_ty)
-        (HasArgType els arg_ty)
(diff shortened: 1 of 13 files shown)


dag_in_context/src/optimizations/conditional_push_in.egg
```diff
@@ -0,0 +1,161 @@
...
+        ; Union the original input with Bop(c, x) in the new regions
+        (union (Get arg_tr i) (Bop o (Const c new_ins_ty if_tr) (Get arg_tr orig_ins_len)))
+        (union (Get arg_fa i) (Bop o (Const c new_ins_ty if_fa) (Get arg_fa orig_ins_len)))
+
+        ; Union the ifs
+        (union if_e (If pred new_ins new_thn new_els))
+      )
+      :ruleset push-in)
+
+(rule (
```
oflatt, line 43: Shouldn't interval analysis handle pushing constants in already?

oflatt (approved):
awesome- I think we should delete the constant rule though

PR #714, opened 2025-01-28
Adding `Bitand`, `Neg`, lowbit tests, and lowbit optimizations

FTRobbin:
This PR adds two new operators `Bitand` and `Neg`, some tests, and analysis and optimization in Egglog that rewrite the naive loop implementation of the lowbit function into the Hacker's Delight one-liner.

This now outperforms llvm-O3-O3 by an order of magnitude on my machine because the closed form is an O(log n) speedup (n = 2 * 10^5).

The RVSDGs before and after the optimization:
lowbit_naive_br-rvsdg-conversion
lowbit_naive_br-rvsdg-optimize


oflatt (approved):
Looks great! Did you really want all of those to be benchmarks? Perhaps the fenwick tree should be a benchmark while your tests should be in `tests/passing/small`
We do more testing on the tests, including generating snapshots for the `small` folder

oflatt:
Will merge after successful nightly! Being careful right now because it seems like eggcc might have crashed the nightly last night...

dag_in_context/src/optimizations/hackers_delight.egg
```diff
@@ -0,0 +1,73 @@
...
+    (let lowbitn (Bop (Bitand) n (Uop (Neg) n)))
+    (union (Get outerif j) lowbitn)
+    (union (Get outerif i) (Bop (Div) n lowbitn))
+) :ruleset hacker)
+
+;; Try to do a state-edge-passthrough for loops
+;; PopcountIterations guarantees termination for non-zero values
+;; lowbit(0) is undefined behavior
+
+(rule (
```
yihozhang, line 65: This technically violates the weak linearity because we have not provided an equivalent value for other values of the then branch?

FTRobbin: My understanding is that if you are computing something else in your loop, you will not be able to extract without the loop.

oflatt: Yihong is right- we need to provide a loop that doesn't have the state edge in it as an alternative. See the existing state edge passthrough file for an example of how it's done for if statements.

If you are confused why we have to do this, hopefully the paper clears it up! Good chance to read that section

oflatt: Also, move this rule to that file and ruleset

oflatt (changes requested):
Main concern is the state edge passthrough rule

dag_in_context/src/optimizations/hackers_delight.egg
FTRobbin: I have fixed this issue, but I feel it makes more sense for the rule to stay in this file.

dag_in_context/src/optimizations/hackers_delight.egg
```diff
@@ -0,0 +1,90 @@
...
+    (NTZIterations anyif n i)
+    (= anyif (If cond inputs thenbr elsebr))
+    (= thenbr (DoWhile lpinputs pred_outputs))
+    (= (Get pred_outputs (+ j 1)) (Get (Arg arg_ty then_ctx) j))
+    (HasType (Get pred_outputs (+ j 1)) (Base (StateT)))
+) (
+    (let newlpinputs (TupleRemoveAt lpinputs j))
+    (let newpred_outputs (TupleRemoveAt pred_outputs (+ j 1)))
+    
+    (let newlpctx (DummyLoopContext newlpinputs newpred_outputs pred_outputs))
```
oflatt, line 77: These days we've decided to use `TmpCtx` instead, then make sure to `delete` it at the end of the rule. It's safer in case you forget to include some important information in the dummy context

oflatt (approved):
Awesome work! Small note about TmpCtx, not blocking but would be nice. Waiting on nightly to merge this

oflatt:
Nightly looks great!

PR #718, opened 2025-01-30
Add simplifycfg pass to llvm output

ezrosent:


oflatt:
Unnecessary turns out, we can try again sometime

PR #719, opened 2025-01-31
Compute the cost of If as the maximum of its branches

yihozhang:


oflatt:
Why max? Should we do average?

yihozhang:
`max` measures the worst case complexity while average only measures the average complexity assuming an equal chance for both branches. If using average, this means we can bring down the cost of an arbitrary expression `e` with `if true e 0`, and so on

oflatt:
Oh yeah, that's a problem with average. 
I'm suspicious of max because it means that one branch is completely ignored- perhaps we should do a hybrid approach? Max plus some some small factor of average

oflatt (approved):

PR #722, opened 2025-02-06
New macros in nightly

oflatt:


ajpal (approved):
I didn't look closely enough to really tell what this is doing, but if you want these macros for writing i trust u

PR #723, opened 2025-02-06
Improve performance of terms ruleset greatly

oflatt:


dag_in_context/src/utility/terms.egg
```diff
@@ -1,43 +1,78 @@
...
+(ruleset terms-helpers)
+(ruleset terms-helpers-helpers)
 
 (sort TermAndCost)
 (function Smaller (TermAndCost TermAndCost) TermAndCost)
 
 (function ExtractedExpr (Expr) TermAndCost
   :merge (Smaller old new))
+;; potential extractions- use so that we don't modify the
+;; timestamp of ExtractedExpr when the costs are equal
```
ajpal, line 11: what timestamp? can you explain this a little more?

oflatt: Added comments! This is egglog's internal timestamp of when this row was changed in the table. Changing it means we'll match on it again.

dag_in_context/src/utility/terms.egg
```diff
@@ -1,43 +1,78 @@
 (ruleset terms)
+(ruleset terms-helpers)
+(ruleset terms-helpers-helpers)
```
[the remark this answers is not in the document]
oflatt: Added a comment- the helpers rely on `Smaller`, which I resolve in another ruleset for a little more performance

ajpal (approved):

PR #735, opened 2025-02-14
Ablate any ruleset

rtjoa:
Remove a ruleset with `--ablate RULESET`

To see a `localnightly.sh` with ablated versions, set the following string in `profile.py` to non-empty (e.g. "loop-inv-motion").
```
TO_ABLATE = "" # change to a ruleset to ablate
```


oflatt (approved):
looks good, hacky but useful

PR #736, opened 2025-02-16
Add extraction time to nightly

yihozhang:
Waiting for the nightly


oflatt (approved):
Looks good, but would be nice to refactor to use a struct instead of a tuple
Easy to write bugs with tuples!

PR #784, opened 2025-11-18
Deadline graph tweaks

oflatt:


dag_in_context/src/tiger/ilp.cpp
```diff
@@ -50,77 +50,28 @@ static void kill_process_group(pid_t pid) {
...
-	}
-	if (WIFEXITED(status)) {
-		return WEXITSTATUS(status);
-	}
-	if (WIFSIGNALED(status)) {
-		return 128 + WTERMSIG(status);
-	}
-	return -1;
-#endif
+static int run_command(const std::string &command) {
```
FTRobbin: Use system(). See:
https://man7.org/linux/man-pages/man3/system.3.html
https://en.cppreference.com/w/cpp/utility/program/system.html

dag_in_context/src/tiger/ilp.cpp
```diff
@@ -780,6 +725,19 @@ Extraction extractRegionILPInner(const EGraph &g, const EClassId root, const vec
 		ofstream out_debug_log("/tmp/tiger_last_extract.log", ios::binary);
 		out_debug_log << in_debug_log.rdbuf();
 	}
+	// cerr << solver_log << endl;
+	bool solver_timed_out = contains_case_insensitive(solver_log, "timeout") || contains_case_insensitive(solver_log, "time limit");
+
+	if (solver_timed_out) {
+		timed_out = true;
+		if (!g_config.time_ilp) {
```
FTRobbin, line 692: Could you clean up this part to just use one stream?

dag_in_context/src/tiger/ilp.cpp
```diff
@@ -632,7 +583,7 @@ Extraction extractRegionILPInner(const EGraph &g, const EClassId root, const vec
 	}
 	lp << "Subject To\n";
 
-	// Require at least one root enode to be picked
+	// Require exactly one root node
 	for (EClassId c = 0; c < (EClassId)g.eclasses.size(); ++c) {
```
FTRobbin: Remove the loop

dag_in_context/src/tiger/ilp.cpp
oflatt: Do you mean by getting rid of the fail function? I'm not sure how to otherwise. I like the fail function because it makes things a little more compact in this function.

FTRobbin: I'm saying that you created an `ostringstream` just to serialize `elapsed_seconds.count()`  into a string to be outputed in `cerr` in `fail`. Seems you can just do one `cerr` call instead?

FTRobbin: You can also use `std::to_string` for `double` floats.

FTRobbin (approved):

PR #788, opened 2026-02-11
Add optimizations to support talk example

ajpal:
```
def foo(x): return (x * 2) / 2

def bar(a, b):
	x = load(a)
set(b, foo(x))
if load(a) != load(b):
	print 100
set(a, foo(load(a))
```

Added peepholes:

- x/x => 1 for constants not equal to 0
- Not false => true + Not true => false
- (a * b) / c => a * (b/c)
- x = x => true

Added memory optimizations (may not be weakly linear but thats okay)
- reorder writes if they don't alias
- load doesn't change state edge (load dropping)

### DIFF
--- a/dag_in_context/src/optimizations/non_weakly_linear.egg
+++ b/dag_in_context/src/optimizations/non_weakly_linear.egg
@@ -68,3 +68,16 @@
   (set (LoopNumItersGuess new-loop-input new-loop-body) (- old_cost 1))
   )
  :ruleset non-weakly-linear)
+
+ ; Two writes can be swapped if they don't alias
+(rule (    
+        (NoAlias w1-addr w2-addr)
+        (= w1 (Top (Write) w1-addr w1-val st))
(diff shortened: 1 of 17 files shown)


dag_in_context/src/optimizations/non_weakly_linear.egg
```diff
@@ -67,4 +67,17 @@
 
   (set (LoopNumItersGuess new-loop-input new-loop-body) (- old_cost 1))
   )
- :ruleset non-weakly-linear)
\ No newline at end of file
+ :ruleset non-weakly-linear)
+
+ ; Two writes can be swapped if they don't alias
+(rule (    
+        (NoAlias w1-addr w2-addr)
```
oflatt, line 74: I do worry this may make the e-graphs a lot bigger, exploring all combinations of writes. We should look at a nightly run

oflatt (commented):
Just waiting on nightly on this one!

PR #789, opened 2026-02-12
Make gurobi optional so nightly runs again

oflatt:
This PR refactors the nightly.sh script, rewriting it in python so we can parse arguments nicely. 
It adds --use-gurobi as a flag that allows you to enable gurobi use. The existing `--paper` flag enables this by default so you can reproduce the paper. 

It also adds --percent-regions to specify how many regions to run on, set to 100 percent for the `--paper` flag. This is necessary because the nightly is too long without it.

### DIFF
--- a/infra/nightly-resources/index.html
+++ b/infra/nightly-resources/index.html
@@ -3,6 +3,7 @@
 <html>
 
 <head>
+  <link rel="icon" href="data:,">
   <script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.8.0/Chart.bundle.min.js"></script>
   <script src="./Plugin.Errorbars.js"></script>
   <script src="chart.js"></script>
--- a/infra/profile.py
+++ b/infra/profile.py
@@ -76,7 +76,7 @@ _base_treatments = [
   "eggcc-WITHCTX-O0-O0",
 ]
 
-# Treatments that require Gurobi (only run in paper mode)
+# Treatments that require Gurobi (when --use-gurobi or --paper is passed)
 _gurobi_treatments = [
   "eggcc-tiger-ILP-O0-O0",
   "eggcc-tiger-ILP-NOMIN-O0-O0",
--- a/infra/profile.py
+++ b/infra/profile.py
@@ -12,33 +12,39 @@ import resource
 import concurrent.futures
 import time
 from generate_cfgs import make_cfgs
+from dataclasses import dataclass
 
 
-# testing mode takes much fewer samples than the real eval in the paper
-IS_TESTING_MODE = True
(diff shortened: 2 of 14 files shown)


infra/nightly-resources/index.html
```diff
@@ -3,6 +3,7 @@
 <html>
 
 <head>
+  <link rel="icon" href="data:,">
```
ajpal, line 6: what does this do?

infra/profile.py
```diff
@@ -12,26 +12,39 @@
...
-    return 6 * 60 * 60 # 6 hours (ILP can take a long time, timing out on lots of regions after 5 min)
+@dataclass
+class NightlyConfig:
+  """Configuration for nightly runs."""
+  paper_mode: bool = False
+  use_gurobi: bool = False
+  
+  @property
+  def eggcc_timeout_secs(self):
+    if not self.paper_mode:
```
ajpal, line 26: nit: (here and elsewhere) prefer positive phrasing for these (`if paper mode then _ else _`)

infra/profile.py
```diff
@@ -43,7 +56,9 @@ def average(lst):
 # use for running a subset of the treatments
 # disables checks that ensure the data is complete
 UNSAFE_TREATMENTS = False
-treatments = [
+
+# Base treatments that don't require Gurobi
+_base_treatments = [
```
ajpal, line 61: why `_` prefixes? Doesn't that usually imply unused vars?

infra/profile.py
```diff
@@ -56,15 +71,33 @@ def average(lst):
...
+  "eggcc-tiger-ILP-NOMIN-O0-O0",
   # run both tiger and ILP on the same egraphs, keep this last in the list
   "eggcc-tiger-ILP-COMPARISON", 
 ]
 
+def get_treatments(config: NightlyConfig):
+  """Returns the list of treatments to run based on the config.
+  When config.use_gurobi is True, includes Gurobi treatments; otherwise only base treatments."""
+  result = _base_treatments.copy()
+  if TO_ABLATE != "":
```
ajpal, line 91: what is `TO_ABLATE`? should it be part of the new config thing?

ajpal (approved):
couple comments/questions, but none are blocking!

infra/nightly-resources/index.html
oflatt: Fixes an annoying warning

infra/profile.py
oflatt: I think the Python convention is that leading `_` means internal, not unused. Not part of the public API. The AI loves to do this

infra/profile.py
oflatt: It's a hack we've had for a while, we can refactor in the future.

PR #792, opened 2026-03-16
Fenwick Revived!

FTRobbin:


dag_in_context/src/optimizations/non_weakly_linear.egg
```diff
@@ -67,4 +67,27 @@
...
+    (= (Get else_ i) (Get (Arg arg_ty else_ctx) j))
+    (HasType (Get then_ i) (Base (StateT)))
+) (
+    (union (Get outputs i) (Get inputs j))
+) :ruleset non-weakly-linear)
+
+; Pass through loop state edge arguments, requires tiger
+; Warning: This potentially eliminate infinite loops
+
+(rule (
```
oflatt, line 87: We should merge this PR after fixing this rule to not eliminate infinite loops. I think we had an analysis for this before? Not sure

oflatt (approved):
nightly looks good

PR #796, opened 2026-06-05
Remove constructors from merge functions

saulshanabrook:
This removes constructor-producing encodings from merge functions so merge results are eager values selected from the existing inputs, which is needed for proof mode and for avoiding merge-driven e-node growth.

## Stack

PR against Oliver's `oflatt-latest-egglog` branch.

This now depends on:

- egglog combined branch: `egraphs-good/egglog`, branch `codex/pr927-plus-pr930-eggcc-proof` (`9b5ce111`), which merges container PR #927 with PR #930 and the proof-mode command-macro type-info fix.
- egglog-experimental helper PR: https://github.com/egraphs-good/egglog-experimental/pull/56 (`19ed866`).

The Cargo manifest points at the experimental PR branch, and `Cargo.lock` pins that branch plus the combined egglog branch.

## What changed

- `TermAndCost` is now a `Pair Term i64`.
- `ExtractedExpr` uses `pair-min-by-second-i64`, which returns one of the existing pair values directly.
- interval `Bound` is now `Maybe (Either i64 bool)`.
- Lower/upper bound merges use `maybe-either-i64-bool-max` and `maybe-either-i64-bool-min`.
- Eggcc-specific aliases like `IntB`, `BoolB`, `Dead`, `bound-int`, and `bound-bool` were removed; eggcc now spells the generic container representation directly.
- `IVTRes` is now `Pair IVTPayload i64`, with merges using `pair-min-by-second-i64`.
- Rules that consume pair-backed values use `pair-first` and `pair-second` instead of matching on pair-shaped LHS terms.
- `--run-mode egglog` emits the generated egglog program for direct use with the experimental CLI.

## Validation

Passed locally:

```sh
cargo check --locked
cargo test -p dag_in_context interval_analysis
cargo test -p dag_in_context optimizations::ivt
cargo test -p dag_in_context optimizations::memory
cargo test -p dag_in_context optimizations::switch_rewrites
cargo run --release -- --run-mode egglog --stop-after-n-passes 1 benchmarks/passing/polybench/linear-algebra/kernels/2mm.bril > .tmp/proof-profile/2mm-pass1-current.egg
/Users/saul/p/wt/egglog-experimental/eggcc-proof-baseline-pr927/target/release/egglog-experimental .tmp/proof-profile/2mm-pass1-current.egg
```

Current composed-stack validation:

```sh
# eggcc, after lockfile repin
cargo check
cargo check --locked

# egglog-experimental PR #56, release CLI against full checked-in 2mm export
target/release/egglog-experimental --proofs tests/fixtures/eggcc-2mm-pass1-merge-old.egg
```

The full 2mm pass-1 export is checked into egglog-experimental PR #56 as `tests/fixtures/eggcc-2mm-pass1-merge-old.egg`; it contains no `:no-merge` and runs successfully in proof mode with the release CLI. The default experimental test suite keeps the full-export canary ignored because debug-profile proof execution was too slow for CI.

## Performance

Earlier local 2mm export runs for the constructor encoding vs eager encoding showed the eager encoding faster (about 13% on that run). I did not rerun the 5x benchmark after replacing eggcc-specific aliases with direct generic container forms.


Cargo.toml
```diff
@@ -63,3 +63,8 @@ write-intermediates = []
 # Runs tests relying on LLVM. This may not be available on all systems.
 llvm = []
 default = ["llvm"]
+
+[patch."https://github.com/saulshanabrook/egg-smol.git"]
```
oflatt: why do we need a different egglog? I thought we were going to make eggcc just add a few primitives using the public egglog API.

dag_in_context/src/utility/terms.egg
```diff
@@ -1,104 +1,94 @@
...
 ;; helpers keeps track of the new best extracted terms
 (ruleset terms-helpers)
-;; helpers-helpers runs `Smaller` rules, resolving the merge function for helpers
+;; Kept as a nested helper ruleset in the existing schedule shape.
 (ruleset terms-helpers-helpers)
 
-(sort TermAndCost)
-(constructor Smaller (TermAndCost TermAndCost) TermAndCost)
+(sort TermAndCost (Pair Term i64))
+(sort UnitThunk (UnstableFn () Unit))
```
oflatt: we discussed removing use of unstablefn for now

oflatt (changes requested):
small comments from our discussion last time

